PR #2, opened 2021-03-23
38417830: ani2psl API call is crushing cNode if the incorrect ANI address is passed

akobrin2:
-added extended error diagnostic for private key decoding (DecodeSecret: used in ani2psl_secret)
	-validate CKey objects after decoding private key string (base58 encoded)
	-show extended error messages for ./pastel-cli ingest ani2psl_secret key
	-replaced all calls to DecodeSecret function to support extended error diagnostic
	-corrected some C++ code to compile on Visual Studio 2019

Pastel Core build improvements:
   -changed required gcc C++ standard from c++11 to c++17 (configure.ac)
   -updated autoconf m4 macros for boost libs,gcc compiler features

pastel-gtest (Google Tests for Pastel Core):
	-added new test_mnode_rpc.cpp - tests for mnode-rpc.cpp, ani2psl_secret parameterized test
	-fixed existing test_block.cpp, added to Makefile.gtest.include
	-implemented command-line options parsing using boost program-options library: (--help,--filter)
	-added support for --filter=xxx option to use Google Test filtering feature (include/exclude tests to run by filter);
    this can be used to debug specific tests much faster

### DIFF
--- a/build-aux/m4/ax_boost_base.m4
+++ b/build-aux/m4/ax_boost_base.m4
@@ -1,5 +1,5 @@
 # ===========================================================================
-#       http://www.gnu.org/software/autoconf-archive/ax_boost_base.html
+#      https://www.gnu.org/software/autoconf-archive/ax_boost_base.html
 # ===========================================================================
 #
 # SYNOPSIS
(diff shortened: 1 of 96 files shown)


a-ok123 (approved):
Merging, as tests run ok. Else, there are mostly clean up

PR #15, opened 2021-05-25
38925962 config file updates

igorzagorodniy:
please review

### DIFF
--- a/src/mnode-config.cpp
+++ b/src/mnode-config.cpp
@@ -58,6 +58,13 @@ std::string get_string(json::iterator& it, std::string name)
         return it->at(name);
     return "";
 }
+
+unsigned int get_uint(json::iterator& it, std::string name)
+{
+    if (it->count(name) && !it->at(name).is_null() && it->at(name).is_number_unsigned())
+        return it->at(name);
+    return -1;
+}
 std::string get_obj_as_string(json::iterator& it, std::string name)
 {
     if (it->count(name) && !it->at(name).is_null() && it->at(name).is_object())
@@ -79,7 +86,7 @@ bool CMasternodeConfig::read(std::string& strErr)
                 {"mnAddress", ""},
                 {"mnPrivKey", ""},
                 {"txid", ""},
-                {"outIndex", ""},
+                {"outIndex", 0},
                 {"extAddress", ""},
                 {"extKey", ""},
                 {"extCfg", {}}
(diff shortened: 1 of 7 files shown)


src/mnode-config.cpp
```diff
@@ -58,6 +58,13 @@ std::string get_string(json::iterator& it, std::string name)
         return it->at(name);
     return "";
 }
+
+unsigned int get_uint(json::iterator& it, std::string name)
```
akobrin1, line 62: pass name parameter by const reference (const std::string &name). Here the copy of the "name" string is created.
Also, you are calling it->at(name) 3 times here.... check it->count separately,  if it is = 0, return.
Then just const auto &value = it->at(name);-> check it and return the value.
It would be good if for all new code we add some comments (you can use doxygen comments style).

src/mnode-masternode.cpp
```diff
@@ -429,7 +429,7 @@ bool CMasternode::VerifyCollateral(CollateralStatus& collateralStatus)
 }
 
 #ifdef ENABLE_WALLET
-bool CMasternodeBroadcast::Create(std::string strService, std::string strKeyMasternode, std::string strTxHash, std::string strOutputIndex, 
+bool CMasternodeBroadcast::Create(std::string strService, std::string strKeyMasternode, std::string strTxHash, unsigned int nOutputIndex,
```
akobrin1, line 432: function is not changing nOutputIndex, just pass it as const unsigned int.

src/mnode-rpc.cpp
```diff
@@ -496,7 +499,7 @@ UniValue masternode(const UniValue& params, bool fHelp)
             mnObj.pushKV("address", mne.getIp());
             mnObj.pushKV("privateKey", mne.getPrivKey());
             mnObj.pushKV("txHash", mne.getTxHash());
-            mnObj.pushKV("outputIndex", mne.getOutputIndex());
+            mnObj.pushKV("outputIndex", boost::lexical_cast<uint64_t>(mne.getOutputIndex()));
```
akobrin1, line 502: we're trying to get rid of boost. You can just use static_cast<uint64_t> here.

src/mnode-validation.cpp
```diff
@@ -45,7 +45,7 @@ int GetUTXOConfirmations(const COutPoint& outpoint)
 }
 
 #ifdef ENABLE_WALLET
-bool GetMasternodeOutpointAndKeys(CWallet* pWalletMain, COutPoint& outpointRet, CPubKey& pubKeyRet, CKey& keyRet, std::string strTxHash, std::string strOutputIndex)
+bool GetMasternodeOutpointAndKeys(CWallet* pWalletMain, COutPoint& outpointRet, CPubKey& pubKeyRet, CKey& keyRet, std::string strTxHash, unsigned int nOutputIndex)
```
akobrin1, line 48: const

src/mnode-validation.h
```diff
@@ -22,7 +22,7 @@ int GetUTXOConfirmations(const COutPoint& outpoint);
 void FillOtherBlockPayments(CMutableTransaction& txNew, int nBlockHeight, CAmount blockReward, CTxOut& txoutMasternodeRet, CTxOut& txoutGovernanceRet);
 
 #ifdef ENABLE_WALLET
-    bool GetMasternodeOutpointAndKeys(CWallet* pwalletMain, COutPoint& outpointRet, CPubKey& pubKeyRet, CKey& keyRet, std::string strTxHash, std::string strOutputIndex);
+    bool GetMasternodeOutpointAndKeys(CWallet* pwalletMain, COutPoint& outpointRet, CPubKey& pubKeyRet, CKey& keyRet, std::string strTxHash, unsigned int nOutputIndex);
```
akobrin1, line 25: const

akobrin1 (changes requested):

a-ok123:
bogus

PR #22, opened 2021-06-04
Pastel/pvs fix bug v501 v606

tsimafeisetchinava000:
Fix high level warnings on PVS review. (**V501 ~ V606**).


a-ok123 (dismissed):

src/gtest/test_joinsplit.cpp
```diff
@@ -579,7 +579,7 @@ TEST(joinsplit, note_plaintexts)
 
     // Check memo() returns by reference, not return by value, for use cases such as:
     // std::string data(plaintext.memo().begin(), plaintext.memo().end());
-    ASSERT_TRUE(decrypted.memo().data() == decrypted.memo().data());
+    ASSERT_TRUE(decrypted.memo().data() == decrypted.memo().data()); //-V501
```
vadymeng0, line 582: This assert is strange, I think it shouldbe:
ASSERT_TRUE(decrypted.memo().data() == note_pt.memo().data());

src/mnode/ticket-processor.cpp
```diff
@@ -750,7 +750,7 @@ bool CPastelTicketProcessor::CreateP2FMSTransactionWithExtra(const CDataStream&
     int fake_key_size = 33;
     size_t non_padded_size = input_bytes.size();
     size_t padding_size = fake_key_size - (non_padded_size % fake_key_size);
-    if (padding_size != 0) {
+    if (padding_size != 0) { //-V547 false warning
```
vadymeng0, line 753: In case that (non_padded_size % fake_key_size) == 0. A redundant chunk with 33 bytes will be insert. Should change the implementation.

src/util.h
```diff
@@ -269,6 +269,6 @@ class InsecureRand
     {
         nRz = 36969 * (nRz & 65535) + (nRz >> 16);
         nRw = 18000 * (nRw & 65535) + (nRw >> 16);
-        return ((nRw << 16) + nRz) % nMax;
+        return ((uint32_t)(nRw << 16) + nRz) % nMax;
```
vadymeng0, line 272: I think it should be fixed to:
return (((uint32_t)nRw << 16) + nRz) % nMax;

tsimafeisetchinava000: what is difference between (uint32_t)(nRw << 16) and ((uint32_t)nRw << 16)?

src/gtest/test_joinsplit.cpp
tsimafeisetchinava000: ASSERT_TRUE(decrypted.memo().data() == note_pt.memo().data());
It is false.
data() is a pointer.
so it is different.

src/mnode/ticket-processor.cpp
tsimafeisetchinava000: so do you want to change this as if(non_padded_size % fake_key_size != 0)?

src/gtest/test_joinsplit.cpp
tsimafeisetchinava000: So it was checked before.
ASSERT_TRUE(decrypted.memo() == note_pt.memo());

src/mnode/ticket-processor.cpp
vadymeng0: Yes I believe that is the best fix.

src/gtest/test_joinsplit.cpp
vadymeng0: I see. That was my bad for not checking what data() returns.

src/util.h
vadymeng0: The best fix should be: return (((int64_t)nRw << 16) + nRz) % nMax;
The difference is:
    -  (uint32_t)(nRw << 16) : Cast after shift
    -  ((uint32_t)nRw << 16): Cast before shift
According to this: https://pvs-studio.com/en/docs/warnings/v629/ 
We must make sure the type of the left argument is 64-bit before the shift, so we must cast it before the sift, that is what I mean.

tsimafeisetchinava000: but nRw is also uint32_t type.
so, it is same.

vadymeng0: nRw is uint32_t type.
But (int64_t)nRw is int64_t type.

Let's do a test:
image

Dicklesworthstone: Good catch here Vadym

src/gtest/test_joinsplit.cpp
akobrin1: They were checking here that decrypted.memo() is a reference.
There is a better way to check if it memo() points to reference using c++17:
ASSERT_TRUE(std::is_lvalue_reference<decltype(decrypted.memo())>::value);

So, decltype - we are getting data type from decrypted.memo(), then checking if it is lvalue reference.

Dicklesworthstone: Nice, cool trick

src/util.h
akobrin1: don't use c-style casts, ((static_cast<int64_t>(nRw) << 16) + nRz) % nMax;

akobrin1 (changes requested):

a-ok123 (dismissed):

a-ok123 (approved):

akobrin1 (approved):

PR #23, opened 2021-06-05
40318112: fix issues with data type change to size_t

akobrin1:
-fixed an issue in CMasternodeBlockPayees::GetBestPayee with payee.GetVoteCount data type change to size_t
-fixed masternodeManager.size() serialization as uint32_t


a-ok123 (approved):
I am approving it now, but next time DO provide all test results, including rpc_mn

PR #30, opened 2021-06-11
ASCII85 encode/decode support integration

devcareer0:
- Integration ascii85 encode / decode functions from https://github.com/dcurrie/ascii85
- Functions not yet called, but integrated, tested & validated via development tests.

Usage: same as i.e.:EncodeBasse64() or DecodeBase64(), with different name but identical return values.

Testing:
- For development test, a special mnode-rpc.cpp was created where mn_bugs.py script call `"masternode - extra" ` is extended with a functionality of printing results via LogPrintf() to debug.log. (Node0's debug.log attached for such results, just search for ASCII85_tests)
- Cross-checked with output results of dcurrie/ascii85 and identical.
[debug.log](https://github.com/pastelnetwork/pastel/files/6639048/debug.log)
(url)


a-ok123 (changes requested):
It is actually better to add gtest to test it

src/utilstrencodings.cpp
```diff
@@ -123,6 +125,75 @@ vector<unsigned char> ParseHex(const string& str)
...
+        int32_t olen = ascii85_get_max_encoded_length(isz);
+        
+        if (olen < 0)
+        {
+            return retVal; //Encode size error
+        }
+        else
+        {
+            uint8_t obuf[olen];
+            olen = encode_ascii85((uint8_t *)istr, isz, obuf, olen);
```
akobrin1, line 144: don't use c-style casts in cpp code. 
reinterpret_cast<const uint8_t *>(istr), and don't forget about const.

src/utilstrencodings.cpp
```diff
@@ -123,6 +125,75 @@ vector<unsigned char> ParseHex(const string& str)
...
+        {
+            uint8_t dbuf[olen];
+            int32_t ilen = decode_ascii85((uint8_t *)ostr, osz, dbuf, olen);
+
+            if (ilen < 0)
+            {
+                *pfInvalid = true;//Decode error
+                return retVal;
+            }
+            retVal = std::vector<unsigned char>(dbuf, dbuf + ilen);
```
akobrin1, line 184: this is again a double copy. Here you creating vector object from dbuf, and then making a copy to retVal. 
just use retVal.assign(dbuf, dbuf + ilen);

src/utilstrencodings.h
```diff
@@ -44,6 +44,10 @@ std::vector<unsigned char> ParseHex(const char* psz);
 std::vector<unsigned char> ParseHex(const std::string& str);
 signed char HexDigit(char c);
 bool IsHex(const std::string& str);
+std::vector<unsigned char> DecodeAscii85(const char* p, bool* pfInvalid = NULL);
```
akobrin1, line 47: nullptr

src/utilstrencodings.h
```diff
@@ -44,6 +44,10 @@ std::vector<unsigned char> ParseHex(const char* psz);
...
+std::string DecodeAscii85(const std::string& str);
+std::string EncodeAscii85(const unsigned char* pch, size_t len);
+std::string EncodeAscii85(const std::string& str);
 std::vector<unsigned char> DecodeBase64(const char* p, bool* pfInvalid = NULL);
```
akobrin1, line 51: nullptr
Also, if functions are not throwing exceptions, but rather returns error code (like in your case) - mark it noexcept. Compiler will generate much faster code in this case.

src/utilstrencodings.cpp
```diff
@@ -123,6 +125,75 @@ vector<unsigned char> ParseHex(const string& str)
...
+
+string DecodeAscii85(const string& str)
+{
+    vector<unsigned char> vchRet = DecodeAscii85(str.c_str());
+    return (vchRet.size() == 0) ? string() : string((const char*)&vchRet[0], vchRet.size());
+}
+
 string EncodeBase64(const unsigned char* pch, size_t len)
 {
     static const char *pbase64 = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
```
akobrin1, line 199: use constexpr for all const variables - evaluated at compile-time, not run-time.
static auto pbase64 = "....

akobrin1 (changes requested):

src/utilstrencodings.cpp
devcareer0: Rest comment is fine and thank you for the valuable information, i learn day by day :) . But this one i might not get the point. You mean put the 0 from the  `'size() == 0' `comparison to `static const int zero_size = 0` or you mean the whole return value shall be in `constexpr()` ?

a-ok123 (approved):

src/utilstrencodings.cpp
akobrin1: No, look. If you declare any constant now in modern c++, you don't have to do it the old way const char * szStr = "my string"; This was evaluated by compiler at run-time. Now for constant strings you can use constexpr auto szStr = "my string"; In this case it is evaluated at compile-time. Compiler automatically here defines type as const char *. Also, for any other static const variables and defines you can also use constexpr. Instead of #define MYVAR 5, you can say constexpr size_t MYVAR = 5; Benefits are again - compile time and definite var type. constexpr can be used for much more like functions, structures etc...

devcareer0: Most probably I'm stupid AF, but i still do not get the point - in this code snippet you highlighted here, what shall i constexpr ? :(  
(Every decode function in that file looks like exactly the same.)

src/gtest/test_mnode_rpc.cpp
```diff
@@ -86,3 +90,32 @@ INSTANTIATE_TEST_SUITE_P(mnode_rpc_ani2psl, PTest_ani2psl_secret,
         }, true, true)
 	));
 
+TEST(ASCII85_Encode_Decode, ASCII85_Encode_Decode) {
```
akobrin1, line 93: Test ASCII85_Encode_Decode is a perfect candidate for parameterized test.
You declare fixture class
class MyTestClass : public ::testing::TestWithParam<....>
and in your case you can use tuple<string, string>  as a parameter
then instead of TEST() - you write TEST_P()
get<0>(GetParam()) -  this is how you access parameters
So you need to write test only once for one comparison.
Then you declare parameterized test definition 
INSTANTIATE_TEST_SUITE_P(..., ....,
   ::testing::Values(
          make_tuple("encoded_string1", "decoded_string1"), 
          make_tuple("encoded_string2", "decoded_string2"), 
          .....

In tests like this it is much easier to add later on any additional special test cases.
It will be just one extra line in INSTANTIATE_TEST_SUITE_P.

src/ascii85.cpp
```diff
@@ -93,7 +93,7 @@ int32_t encode_ascii85 (const uint8_t *inp, int32_t in_length, uint8_t *outp, in
     if (out_length < 0)
     {
         // At this point altough ascii85_get_max_encoded_length() already returned an error.
-        LogPrintf("ERROR(encode:ascii85) Lenght smaller then zero %d", out_length);
+        LogPrintf("ERROR(encode:ascii85) Length smaller then zero %d", out_length);
```
akobrin1, line 96: still then -> than :)

src/utilstrencodings.cpp
```diff
@@ -125,73 +125,77 @@ vector<unsigned char> ParseHex(const string& str)
...
     return retVal;
 }
 
-string DecodeAscii85(const string& str)
+string DecodeAscii85(const string& str) noexcept
 {
     vector<unsigned char> vchRet = DecodeAscii85(str.c_str());
-    return (vchRet.size() == 0) ? string() : string((const char*)&vchRet[0], vchRet.size());
+    static constexpr int zero_size = 0;
+    return (vchRet.size() == zero_size) ? string() : string((const char*)&vchRet[0], vchRet.size());
```
akobrin1, line 200: I was writing about constexpr for another line of code (about pbase64).
Of course, here you don't need to add any constexpr.  
Also, to check vector for 0-size you could use vchRet.empty() 
return (vchRet.empty()) ? string() : string((const char*)&vchRet[0], vchRet.size());

src/utilstrencodings.cpp
```diff
@@ -125,73 +125,77 @@ vector<unsigned char> ParseHex(const string& str)
...
 
-            if (ilen < 0)
-            {
-                *pfInvalid = true;//Decode error
-                return retVal;
-            }
-            retVal = std::vector<unsigned char>(dbuf, dbuf + ilen);     
+        if (ilen < 0)
+        {
+            *pfInvalid = true;//Decode error
```
akobrin1: still no check for pfInvalid

src/utilstrencodings.cpp
```diff
@@ -125,73 +125,77 @@ vector<unsigned char> ParseHex(const string& str)
...
-        {
-            uint8_t obuf[olen];
-            olen = encode_ascii85((uint8_t *)istr, isz, obuf, olen);
-            if(0 < olen)
-            {
-                std::ostringstream ss;
-                std::copy(obuf, obuf+olen, std::ostream_iterator<uint8_t>(ss, ""));
-                retVal = ss.str();
-            }
+            std::ostringstream ss;
```
akobrin1, line 148: double copy not fixed

src/utilstrencodings.cpp
```diff
@@ -123,6 +125,75 @@ vector<unsigned char> ParseHex(const string& str)
...
+                retVal = ss.str();
+            }
+        }
+    }
+    return retVal;
+}
+
+string EncodeAscii85(const string& str)
+{
+    return EncodeAscii85((const char*)str.c_str(), str.size());
```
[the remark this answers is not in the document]
akobrin1: don't see you removed (const char *)

akobrin1 (changes requested):

a-ok123 (approved):

src/utilstrencodings.cpp
devcareer0: It is. You were checking 1 commit behind!

src/utilstrencodings.cpp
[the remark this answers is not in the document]
devcareer0: It is. You were checking 1 commit behind!

a-ok123 (approved):

src/gtest/test_mnode_rpc.cpp
```diff
@@ -86,3 +90,30 @@ INSTANTIATE_TEST_SUITE_P(mnode_rpc_ani2psl, PTest_ani2psl_secret,
...
+    static void SetUpTestCase()
+    {
+        SelectParams(CBaseChainParams::Network::REGTEST);
+    }
+};
+
+
+TEST_P(PTest_ASCII85_Encode_Decode, ASCII85_Encode_Decode) {
+
+    const string to_be_encoded_1 = get<0>(GetParam()); //"hello"; // Encoded shall be : "BOu!rDZ"
```
akobrin1, line 105: Yes, thank you. It is much easier to add new special test cases to parameterized tests.
Especially for any encode/decode, encrypt/decrypt methods testing.

akobrin1 (approved):

PR #40, opened 2021-06-22
Username Change Request Ticket implementation

vadymeng0:
Implement: https://freedcamp.com/view/2956206/tasks/38977766 
Result:
- Built successfully
- Passed all below test suite on local machine:
cd qa/test-suite
	./mn_main.py
	./mn_payment.py
	./mn_tickets.py
	./mn_tickets_validation.py
	./mn_messaging.py
        ./full_test_suite.py btest
        ./full_test_suite.py gtest
        ./full_test_suite.py sec-hard
        ./full_test_suite.py no-dot-so
        ./full_test_suite.py util-test
        ./full_test_suite.py secp256k1
        ./full_test_suite.py libsnark
        ./full_test_suite.py univalue
        ./full_test_suite.py rpc-common
        ./full_test_suite.py rpc-ext
Test case that is failing:       
	./mn_governance.py  <-- Failed on master branch also.

Test case that took too long to complete, and failed some test cases while running:
./full_test_suite.py rpc-mn


src/mnode/mnode-controller.cpp
```diff
@@ -33,6 +33,8 @@ void CMasterNodeController::SetParameters()
     MasternodeFeePerMBDefault           = 50;
     ArtTicketFeePerKBDefault           = 3;
 
+    MasternodeUsernameFirstChangeFree   = 100;
```
akobrin1, line 36: FirstChangeFee?

src/mnode/mnode-controller.h
```diff
@@ -60,6 +60,8 @@ class CMasterNodeController
     CAmount MasternodeFeePerMBDefault;
     CAmount ArtTicketFeePerKBDefault;
 
+    CAmount MasternodeUsernameFirstChangeFree;
+    CAmount MasternodeUsernameChangeAgainFree;
```
akobrin1, line 64: ChangeAgainFee, it's not free :)

akobrin1 (changes requested):
Fix fee var typo

src/mnode/mnode-controller.cpp
vadymeng0: thanks for correcting it, will fix

src/mnode/mnode-controller.h
vadymeng0: thanks for correcting it, will fix

akobrin2 (approved):

src/mnode/mnode-pastel.cpp
```diff
@@ -1571,3 +1571,175 @@ bool CTakeDownTicket::FindTicketInDb(const std::string& key, CTakeDownTicket& ti
...
 }
+
+
+// CChangeUsernameTicket ///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+CChangeUsernameTicket CChangeUsernameTicket::Create(std::string _pastelID, std::string _username, const SecureString& strKeyPass)
+{
+    CChangeUsernameTicket ticket(std::move(_pastelID));
+
+    ticket.username = _username;
```
VitaliyVK777, line 1582: use std::move

src/mnode/mnode-pastel.cpp
```diff
@@ -1571,3 +1571,175 @@ bool CTakeDownTicket::FindTicketInDb(const std::string& key, CTakeDownTicket& ti
...
+
+    // Check if PastelID already have a username on the blockchain. 
+    if (!masterNodeCtrl.masternodeTickets.FindTicketBySecondaryKey(ticket)) {
+        // IF PastelID has no Username yet, the fee is 100 PSL
```
VitaliyVK777, line 1586: it has 100 PSL by default, enough to set 5000 only if need?

src/mnode/mnode-pastel.cpp
```diff
@@ -1571,3 +1571,175 @@ bool CTakeDownTicket::FindTicketInDb(const std::string& key, CTakeDownTicket& ti
...
+
+        if (pwalletMain->GetBalance() < fullTicketPrice * COIN) {
+            throw std::runtime_error(strprintf("Not enough coins to cover price [%" PRId64 "]", fullTicketPrice));
+        }
+    }
+
+    // Check if username is a bad username. For now check if it is empty only.
+    std::string badUsernameError;
+    if (isUsernameBad(username, badUsernameError)) {
+        throw std::runtime_error(strprintf(badUsernameError.c_str()));
```
VitaliyVK777, line 1660: throw std::runtime_error(badUsernameError);

src/mnode/mnode-pastel.cpp
```diff
@@ -1571,3 +1571,175 @@ bool CTakeDownTicket::FindTicketInDb(const std::string& key, CTakeDownTicket& ti
...
+                           "[this ticket block = %u txid = %s; found ticket block  = %u txid = %s]",
+                           username, existingTicket.GetBlock(), existingTicket.m_txid, m_nBlock, m_txid));
+    }
+
+    // D. Check if this PastelID hasn't changed Username in last 24 hours.
+    CChangeUsernameTicket _ticket;
+    _ticket.pastelID = pastelID;
+    if (masterNodeCtrl.masternodeTickets.FindTicketBySecondaryKey(_ticket)) {
+        const unsigned int height = (preReg || IsBlock(0)) ? chainHeight : m_nBlock;
+        if (height - _ticket.m_nBlock <= 24 * 24) {
```
VitaliyVK777, line 1688: height is unsigned int. I would write here a bit another like
if (height + 24*24 > _ticket.m_nBlock) { ...

qa/rpc-tests/mn_tickets.py
```diff
@@ -1865,6 +1866,88 @@ def takedown_ticket_tests(self):
...
+        assert_equal(non_mn1_bad_username_registered["validationError"], "Username is not valid, it is already registered")
+
+        # good username
+        non_mn1_bad_username_good = self.nodes[self.non_mn4].tickets("tools", "validateusername", "goodname")
+        print(non_mn1_bad_username_good)
+        assert_equal(non_mn1_bad_username_good["isBad"], False)
+        assert_equal(non_mn1_bad_username_good["validationError"], "")
+
+        # Change previous name to new name.
+        try:
```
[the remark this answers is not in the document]
VitaliyVK777: You have limit to change user name next time after 24 hours? why does it work then as here is missing 24*24 blocks generated before.
Have you 24 hours tests?

src/mnode/mnode-pastel.cpp
```diff
@@ -1571,3 +1571,175 @@ bool CTakeDownTicket::FindTicketInDb(const std::string& key, CTakeDownTicket& ti
...
+            // D.2 IF PastelID has changed Username in last 24 hours (~24*24 blocks), do not allow them to change
+            throw std::runtime_error(strprintf("%s ticket is invalid. Already changed in last 24 hours. PastelID - [%s]", GetTicketDescription(TicketID::Username), pastelID));
+        }
+    }
+
+    // E. Check if ticket fee is valid
+    if (!masterNodeCtrl.masternodeTickets.FindTicketBySecondaryKey(_ticket)) {
```
VitaliyVK777, line 1695: you have called FindTicketBySecondaryKey before

src/mnode/mnode-rpc.cpp
```diff
@@ -2320,6 +2369,27 @@ As json rpc
                 mnObj.pushKV("totalstoragefee", totalFee);
                 return mnObj;
             }
+            case RPC_CMD_LIST::validateusername: {
+                std::string username;
+                if (params.size() > 2) {
+                    username = params[2].get_str();
```
VitaliyVK777, line 2375: declare 'const std::string username' here

qa/rpc-tests/mn_tickets.py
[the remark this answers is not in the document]
VitaliyVK777: Any tests when user1 changed the 'name1' to 'name2' and another user2 wants to set his username to 'name1'? I suppose it will not work as you have throw in IsValid if 'name1' is found in db. But it should work as 'name1' is free.

src/mnode/mnode-pastel.cpp
```diff
@@ -1571,3 +1571,175 @@ bool CTakeDownTicket::FindTicketInDb(const std::string& key, CTakeDownTicket& ti
...
+    if (!CPastelID::Verify(reinterpret_cast<const unsigned char *>(strThisTicket.c_str()), strThisTicket.size(),
+                           signature.data(), signature.size(),
+                           pastelID))
+    {
+        throw std::runtime_error(strprintf("%s ticket's signature is invalid. PastelID - [%s]", GetTicketDescription(TicketID::Username), pastelID));
+    }
+    
+    // C (ticket transaction replay attack protection)
+    CChangeUsernameTicket existingTicket;
+    if ((FindTicketInDb(username, existingTicket)) &&
```
VitaliyVK777, line 1675: it does not duplicate CheckTicketExist above? How it will work if user1 changed the 'name1' to 'name2' and another user2 wants to set his username to 'name1'?

src/mnode/mnode-rpc.cpp
```diff
@@ -1981,12 +1981,50 @@ As json rpc
...
+)" + HelpExampleRpc("tickets",
+                    R"("register", "username", "jXYqZNPj21RVnwxnEJ654wEdzi7GZTZ5LAdiotBmPrF7pDMkpX1JegDMQZX55WZLkvy9fxNpZcbBJuE8QYUqBF", "bsmith84", "passphrase")"));
+            std::string pastelID = params[2].get_str();
+            std::string username = params[3].get_str();
+            SecureString strKeyPass;
+            strKeyPass.reserve(100);
+            strKeyPass = params[4].get_str().c_str();
+            CChangeUsernameTicket changeUsernameTicket = CChangeUsernameTicket::Create(pastelID, username, strKeyPass);
+            std::string txid = CPastelTicketProcessor::SendTicket(changeUsernameTicket);
+            mnObj.pushKV(RPC_KEY_TXID, txid);
```
VitaliyVK777, line 2020: std::move(txid)

VitaliyVK777 (changes requested):

src/mnode/mnode-pastel.h
```diff
@@ -618,3 +619,1814 @@ class CTakeDownTicket : public CPastelTicket
...
+    CChangeUsernameTicket() = default;
+
+	explicit CChangeUsernameTicket(std::string _pastelID) :
+        pastelID(std::move(_pastelID))
+    {}
+    std::string KeyOne() const noexcept override { return username; }
+    std::string KeyTwo() const noexcept override { return pastelID; }
+
+    bool HasKeyTwo() const noexcept override { return true; }
+    bool HasMVKeyOne() const noexcept override { return false; }
```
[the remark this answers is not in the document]
vadymeng0: It is needed as a good practice.
**override** keyword served two purposes:
1. It shows the reader of the code that "this is a virtual method, that is overriding a virtual method of the base class."
2. The compiler also knows that it's an override, so it can "check" that you are not altering/adding new methods that you think are overrides.
So, place it here makes the code easier to read.
We can use "final" instead of "override". Yes. But it might leads to another unwanted things. For example,  some unit tests want to mock this class to test, and by defining the function as final, we can't inherit the function again to do that.
For conclusion, using "override" is the best way I believe.

src/mnode/mnode-pastel.h
```diff
@@ -618,3 +619,1814 @@ class CTakeDownTicket : public CPastelTicket
...
+    bool HasMVKeyTwo() const noexcept override { return false; }
```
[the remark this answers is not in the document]
vadymeng0: It is needed as a good practice.
override keyword served two purposes:

It shows the reader of the code that "this is a virtual method, that is overriding a virtual method of the base class."
The compiler also knows that it's an override, so it can "check" that you are not altering/adding new methods that you think are overrides.
So, place it here makes the code easier to read.
We can use "final" instead of "override". Yes. But it might leads to another unwanted things. For example, some unit tests want to mock this class to test, and by defining the function as final, we can't inherit the function again to do that.
For conclusion, using "override" is the best way I believe.

src/mnode/mnode-consts.h
```diff
@@ -15,6 +15,7 @@ constexpr auto TICKET_NAME_ART_BUY      = "art-buy";   // art buy ticket
 constexpr auto TICKET_NAME_ART_TRADE    = "art-trade"; // art trade ticket
 constexpr auto TICKET_NAME_TAKE_DOWN    = "take-down";
 constexpr auto TICKET_NAME_ART_ROYALTY  = "art-royalty"; // art royalty ticket
+constexpr auto TICKET_NAME_USERNAME_CHANGE    = "username-change";
```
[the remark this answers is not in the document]
vadymeng0: The ticket name is "Username Change", the shorter name will make its less meaning

src/mnode/mnode-pastel.h
```diff
@@ -618,3 +619,1814 @@ class CTakeDownTicket : public CPastelTicket
...
+
+    void SetKeyOne(std::string val) override { username = std::move(val); }
+
+    static CChangeUsernameTicket Create(std::string _pastelID, std::string _username, const SecureString& strKeyPass);
```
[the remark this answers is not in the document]
vadymeng0: will fix

src/mnode/mnode-pastel.h
```diff
@@ -618,3 +619,1814 @@ class CTakeDownTicket : public CPastelTicket
...
+public:
+    std::string pastelID;
+    std::string username;
+    CAmount fee{100};
+    std::vector<unsigned char> signature;
+
+public:
+    CChangeUsernameTicket() = default;
+
+	explicit CChangeUsernameTicket(std::string _pastelID) :
```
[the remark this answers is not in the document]
vadymeng0: will fix

qa/rpc-tests/mn_tickets.py
```diff
@@ -1865,6 +1866,88 @@ def takedown_ticket_tests(self):
...
+
+        try:
+            tickets_username_txid1 = self.nodes[self.non_mn3].tickets("register", "username",
+                                                        self.artist_pastelid1, "bsmith84", "passphrase")
+            self.__wait_for_ticket_tnx()
+            mn0_ticket_username_1 = self.nodes[self.non_mn3].tickets("get", tickets_username_txid1["txid"])
+            print(mn0_ticket_username_1)
+
+            self.__wait_for_ticket_tnx()
+            assert_equal(mn0_ticket_username_1["ticket"]["pastelID"], self.artist_pastelid1)
```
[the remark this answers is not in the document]
vadymeng0: will fix

qa/rpc-tests/mn_tickets.py
```diff
@@ -1865,6 +1866,88 @@ def takedown_ticket_tests(self):
         # ...
         print("Take down tickets tested")
 
+    # ===============================================================================================================
+    def username_ticket_tests(self):
+        print("== Username Tickets test ==")
+        # Register first time
+
+        try:
```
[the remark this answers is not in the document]
vadymeng0: we need try/catch to catch the error and assert if it is the expected error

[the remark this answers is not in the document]
vadymeng0: Yes, it is not needed if the test ran as expected - no error shows up.
But what if someone implementing something that affecting current tests? This try/catch will save us 1 re-run time(which is taking like 2 hours) to check what is the error.
Since this is just test cases file. I don't see the reason why we have to save try/catch here.

src/mnode/mnode-pastel.h
```diff
@@ -618,3 +619,1814 @@ class CTakeDownTicket : public CPastelTicket
...
+    *     - Doesn't start with letters.
+    *     - Username registered on the blockchain.
+    *     - Contains bad words (swear, racist,...)
+    * return: true if bad, false if good to use
+    */
+    static bool isUsernameBad(const std::string& username, std::string& error);
+
+};
+
+struct UsernameBadWords {
```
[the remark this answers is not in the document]
vadymeng0: This is created to serve UsernameChange ticket only, so I think it is better to place it here.
Using Singleton is better than just a function with staic list initialized because of **testability**. We can do unit test easily with Singleton, but we can't achieve it easily with a function with static list initialized at first call.

qa/rpc-tests/mn_tickets.py
```diff
@@ -1865,6 +1866,88 @@ def takedown_ticket_tests(self):
...
+            tickets_username_txid1 = self.nodes[self.non_mn3].tickets("register", "username",
```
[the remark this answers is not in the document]
vadymeng0: will fix

qa/rpc-tests/mn_tickets.py
```diff
@@ -1865,6 +1866,88 @@ def takedown_ticket_tests(self):
...
+                                                        self.artist_pastelid1, "bsmith84", "passphrase")
+            print(tickets_username_txid2)
+            self.__wait_for_ticket_tnx()
+
+        except JSONRPCException as e:
+            self.errorString = e.error['message']
+        assert_equal(self.errorString , "Ticket (username-change) is invalid - This Username is already registered in blockchain [Username = bsmith84]")
+
+        # length not valid
+        non_mn1_bad_username_length = self.nodes[self.non_mn4].tickets("tools", "validateusername", "abc")
```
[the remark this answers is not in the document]
vadymeng0: will add one more test case for this

qa/rpc-tests/mn_tickets.py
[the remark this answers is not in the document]
vadymeng0: will add test case for this

src/mnode/mnode-pastel.cpp
vadymeng0: 100 psl first time, 5000 second+ time

src/mnode/mnode-pastel.cpp
vadymeng0: will fix

qa/rpc-tests/mn_tickets.py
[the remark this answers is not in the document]
VitaliyVK777: when I see try/catch - I expect to get error here. In "mn_tickets" tests we are trying to separate working code and code with errors.
I thought exception will come any case if exist. But present try/catch for working code looks a bit different from other implementation in the mn_tickets.

src/mnode/mnode-consts.h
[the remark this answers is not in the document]
VitaliyVK777: ok. no problem here. but for me shorter is better ;)

src/mnode/mnode-rpc.cpp
vadymeng0: absolutely shouldn't

VitaliyVK777 (changes requested):

src/mnode/mnode-rpc.cpp
vadymeng0: will fix

src/mnode/mnode-pastel.cpp
vadymeng0: you mean height <= _ticket.m_nBlock + 24*24 ?

src/mnode/mnode-pastel.cpp
vadymeng0: it is allowed as discussed with the team.

src/mnode/mnode-pastel.cpp
VitaliyVK777: yep, and you have set 100 PSL as default value in the header. It is not bug. Just repeated code. But if you want to get it more visible here - then it is fine.

src/mnode/mnode-pastel.cpp
vadymeng0: will fix

VitaliyVK777 (changes requested):

src/mnode/mnode-pastel.cpp
VitaliyVK777: I'm just worry we have 2 repeated calls to database that have the same result.

VitaliyVK777 (changes requested):

src/mnode/mnode-pastel.cpp
vadymeng0: it is set as default value in Create(), this is is to check if if does match with that default value :)

src/mnode/mnode-pastel.cpp
VitaliyVK777: Yep, failed, should be
_ticket.m_nBlock + 24*24 >= height

VitaliyVK777 (changes requested):

src/mnode/mnode-pastel.h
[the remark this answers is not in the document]
vadymeng0: yes I will repeat that 98 function to make to code easier to read.

src/mnode/mnode-pastel.cpp
VitaliyVK777: ok ) ticket.fee=100PSL by default so it could be just next
     if (masterNodeCtrl.masternodeTickets.FindTicketBySecondaryKey(ticket)) {
       ticket.fee = masterNodeCtrl.MasternodeUsernameChangeAgainFee;
     }
but it is not problem to keep current code too

VitaliyVK777 (changes requested):

src/mnode/mnode-pastel.h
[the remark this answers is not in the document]
vadymeng0: especially for important function like these HasKeyOne, HasKeyTwo

src/mnode/mnode-pastel.cpp
vadymeng0: will fix

vadymeng0: will fix

src/mnode/mnode-pastel.cpp
vadymeng0: they are used in different context, different conditions. Please check if `(pReg)`,  and also the conditions of `FindTicketInDb()`

src/mnode/mnode-pastel.cpp
vadymeng0: I do not prefer the idea that we always try to make the code as shorter as possible, while leaving it there make the code easier to understand.

src/mnode/mnode-pastel.cpp
VitaliyVK777: Probably it will be throw on CheckTicketExist and will not come here. I think you need the first to test next: "when user1 changed the 'name1' to 'name2' and another user2 wants to set his username to 'name1'". Probably you will need to do more updates, I did not think about like test case yet. I can think about this if you need help.

VitaliyVK777 (changes requested):

src/mnode/mnode-rpc.cpp
vadymeng0: fixed

qa/rpc-tests/mn_tickets.py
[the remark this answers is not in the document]
vadymeng0: Yes, the exception will come any case if exist. But without try catch, it will stop the test and throw that error.
What I mean is I want to let it run without stopping. In the case I mentioned (implementation changes that affect the test results), multiple errors will print together, so we can check them as 1 run time only.

src/mnode/mnode-pastel.cpp
vadymeng0: that test case is tested already, not sure why it is removed in mn_tickets.py. Will add back.

src/mnode/mnode-pastel.cpp
vadymeng0: fixed

src/mnode/mnode-pastel.cpp
vadymeng0: fixed

src/mnode/mnode-pastel.h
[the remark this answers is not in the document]
vadymeng0: fixed

src/mnode/mnode-pastel.cpp
vadymeng0: fixed

qa/rpc-tests/mn_tickets.py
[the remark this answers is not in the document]
VitaliyVK777: I agree with you that this can save time during testing but after testing it should be removed. Consider that some bug was added to the code after some time - you will be not informed about this new bug with exception as try/catch will hide it! Do you check the log of every call of every test during running different 60 tests? Should another coder to know your tests/log text and if is correct/incorrect?

src/mnode/mnode-pastel.h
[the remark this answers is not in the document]
VitaliyVK777: mnode_pastel.h is very big in length due to this function, need to scroll a lot. Using like functions usualy move to some utils.
you can place the singleton in the unit tests if you need to use it. do you use it now in tests?

VitaliyVK777 (changes requested):

VitaliyVK777:
One solution:
1) add additional bool field "active" to Username ticket
2) FindTicketInDb probably should return you the latest Username ticket
3) check "active" flag
4) when user changing name need to write additional ticket with old user name and "active" false. But I do not like that need to create this second ticket.

Second solution:
1) 'Username' ticket should have 2 fields - old and new name
2) During request to add new username you need to check if exist similar ticket with 'new_name' with the highest block, and check if exist similar ticket with 'old_name' with the highest block. And compare their block height. If 'new_name block height' > 'old_name block height' then this username is used, otherwise if 'old_name block height' > 'new_name block height' then this username was used and free now.

qa/rpc-tests/mn_tickets.py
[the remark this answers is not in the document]
vadymeng0: makes sense, will fix

src/mnode/mnode-pastel.h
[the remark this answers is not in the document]
vadymeng0: Oh so we dont have to scroll util file if we move it to util?
And singleton is better for testabilities but now we didnt write tests for it yet, so we should use a worst version (function with static list). So in the future if we need ot test it we change back to Singleton?

a-ok123 (approved):

src/mnode/mnode-pastel.cpp
```diff
@@ -1571,3 +1571,172 @@ bool CTakeDownTicket::FindTicketInDb(const std::string& key, CTakeDownTicket& ti
...
+    if (!std::all_of(username.begin(), username.end(), [&](unsigned char c) 
+        { 
+          return (upper.find(c) != std::string::npos || lower.find(c) != std::string::npos || digits.find(c) != std::string::npos); 
+        }
+      )) {
+        error = "Invalid username, should contains letters A-Z a-z, or digits 0-9 only";
+        return true;
+    }
+    // Check if contains bad words (swear, racist,...)
+    for (const auto& elem : UsernameBadWords::Singleton().wordSet) {
```
akobrin1, line 1734: what about lowercase/upppercase/mixed case words? 
Make a copy of username and use lowercase() function from str_utils.h. Then you can search words from the badword set - they should be all lowercased as well.

src/mnode/mnode-pastel.h
```diff
@@ -618,3 +619,1812 @@ class CTakeDownTicket : public CPastelTicket
...
+		"signature": ""
+	},
+ */
+class CChangeUsernameTicket : public CPastelTicket
+{
+public:
+    std::string pastelID;
+    std::string username;
+    CAmount fee{100};
+    std::vector<unsigned char> signature;
```
akobrin1, line 639: you can use new type v_uint8 (from vector_type.h)

src/mnode/mnode-pastel.h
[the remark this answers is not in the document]
akobrin1: it would be beneficial also later on to split mnode-pastel.h|cpp into separate cpp|h files for each ticket type.
Much easier to merge

src/mnode/mnode-rpc.cpp
```diff
@@ -2424,6 +2473,27 @@ As json rpc
...
+                    bool isBad = CChangeUsernameTicket::isUsernameBad(username, usernameValidationError);
+                    if (!isBad) {
+                        CChangeUsernameTicket existingTicket;
+                        if (CChangeUsernameTicket::FindTicketInDb(username, existingTicket)) {
+                            isBad = true;
+                            usernameValidationError = "Username is not valid, it is already registered";
+                        }
+                    }
+                    obj.pushKV("isBad", isBad);
+                    obj.pushKV("validationError", usernameValidationError);
```
akobrin1, line 2492: usernameValidationError is not used after that... just move it. obj.pushKV("validationError", move(usernameValidationError));

akobrin1 (changes requested):

src/mnode/mnode-pastel.h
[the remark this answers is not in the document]
vadymeng0: will fix

src/mnode/mnode-pastel.cpp
```diff
@@ -1571,3 +1571,172 @@ bool CTakeDownTicket::FindTicketInDb(const std::string& key, CTakeDownTicket& ti
...
+    std::string digits = "0123456789";
+
+    // Check if has only <4, or has more than 12 characters
+    if ((username.size() < 4) || (username.size() > 12)) {
+        error = "Invalid size of username, the size should have at least 4 characters, and at most 12 characters";
+        return true;
+    }
+
+    // Check if doesn't start with letters.
+    if ( (upper.find(username.front()) == std::string::npos) && (lower.find(username.front()) == std::string::npos ) ) {
```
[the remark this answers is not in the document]
vadymeng0: yes I was going to use isalpha before. But then I realize that the isalpha result will depend on the locale. For example, it returns true for `unsigned char c = '\xdf';` for ISO-8859-1 locale in this [example](https://en.cppreference.com/w/cpp/string/byte/isalpha). 
So I moved to this approach to make it clearer and prevent dealing with locale.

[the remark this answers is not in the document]
akobrin1: can you add then functions to str_utils.h like these:
inline bool isalphaex(const char c) noexcept { return ((c >= 'A') && (c <= 'Z')) || ((c >= 'a') && (c <= 'z')); }
inline bool isdigitex(const char c) noexcept { return (c >= '0') && (c <= '9'); }
and use instead of string search. I wrote small test program to compare two approaches and using isalphaex etc functions is ~8 times faster.

src/mnode/mnode-pastel.cpp
vadymeng0: fixed

src/mnode/mnode-pastel.cpp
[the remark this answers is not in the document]
vadymeng0: fixed

src/mnode/mnode-pastel.h
vadymeng0: fixed

src/mnode/mnode-rpc.cpp
vadymeng0: fixed

src/mnode/mnode-pastel.h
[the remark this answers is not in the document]
vadymeng0: fixed

akobrin1 (dismissed):
looks good

a-ok123 (approved):

akobrin1 (approved):

VitaliyVK777 (approved):

vadymeng0:
> One solution:
> 
> 1. add additional bool field "active" to Username ticket
> 2. FindTicketInDb probably should return you the latest Username ticket
> 3. check "active" flag
> 4. when user changing name need to write additional ticket with old user name and "active" false. But I do not like that need to create this second ticket.
> 
> Second solution:
> 
> 1. 'Username' ticket should have 2 fields - old and new name
> 2. During request to add new username you need to check if exist similar ticket with 'new_name' with the highest block, and check if exist similar ticket with 'old_name' with the highest block. And compare their block height. If 'new_name block height' > 'old_name block height' then this username is used, otherwise if 'old_name block height' > 'new_name block height' then this username was used and free now.

--> No need to use these complex solutions.
leveldb is a key-value database, so everytime a pastelID change their username, the **_keytwo(pastelID) - value (username)_** will change too.
So all we have to do to cover the case that A uses name "usrname1", then change to "usrname2". If B wants to uses name "usrname1":
    + We check what pastelID register this username in the past by using FindTicketInDB.
    + We get the current username that pastelID is using by getting **_keytwo(pastelID) - value (username)_** .
    + If they are match, the username is currently.
    + If they are not match, it means the username sued to be used by that pastelID, but now that pastelID uses another name.
This is implemented already in the PR.

PR #41, opened 2021-06-24
Update circle-ci configurations

vadymeng0:
### What is this PR about?
Change circle-ci config:
    + `xlarge `-->` medium+`
    + -j16 to -j2

## How to test it manually?
Open all 3 build systems: MacOS, Windows, Linux and check if
    + The Executor changed to `Docker Medium+`:
image
    + At the start of "Building Pastel Core" log, check if `-j2` is used.
image

## What is the test result?
This change does not effect source code so no need to run test cases to validate.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ jobs:
   build_and_test_linux:
     executor: pastel-builder
     working_directory: /pastel
-    resource_class: xlarge
+    resource_class: medium+
     steps:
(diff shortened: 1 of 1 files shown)


akobrin1 (approved):
For medium+ we could use -j4 as it should have 4 cpus, but that's ok. We will try to change it in the next check-in.

a-ok123 (approved):

PR #42, opened 2021-06-24
Ownership validation API implementation

devcareer0:
- Implemented Ownership validation logic according to freedcamp task (https://freedcamp.com/dashboard/home/panel/task/40506828)
- Extended mn_tickets.py script with testing against implemented logic. (Tests were running successfully)

### DIFF
--- a/src/mnode/ticket-processor.cpp
+++ b/src/mnode/ticket-processor.cpp
@@ -1087,6 +1087,80 @@ bool CPastelTicketProcessor::ParseP2FMSTransaction(const CMutableTransaction& tx
     return true;
 }
 
+std::vector<std::string> CPastelTicketProcessor::ValidateOwnership(const std::string &_txid, const std::string &_pastelID)
+{
+    //0th: art
+    //1th: trade
+    std::vector<string> sRetVal = {"", ""};
+
+    //Check if ticket is found by txid
+    auto ticket = CPastelTicketProcessor::GetTicket(_txid, TicketID::Art);
+    auto artTicket = dynamic_cast<CArtRegTicket*>(ticket.get());
+    if (!artTicket)
+    {
+        return sRetVal;
+    }
+    else{
+
+        // Check if author and _pastelID are equal
+        if(artTicket->pastelIDs[0].compare(_pastelID) == 0)
+        {
+            sRetVal[0] = _txid;
+        }
+        else
+        {
+            // List all available trade tickets
+            std::vector<CArtTradeTicket> allTickets;
+            listTickets<CArtTradeTicket>([&](const CArtTradeTicket& t)
+            {
+                allTickets.push_back(t);
+            });
+
+            unsigned int chainHeight = 0;
+            {
+                LOCK(cs_main);
+                chainHeight = static_cast<unsigned int>(chainActive.Height()) + 1;
+            }
+            std::vector<CArtTradeTicket> filteredTickets;
+            for (const auto& t : allTickets)
+            {
+                //check if the sell ticket is confirmed
+                if (chainHeight - t.GetBlock() < masterNodeCtrl.MinTicketConfirmations)
+                    continue;
+                //find Trade tickets listing this Trade ticket txid as art ticket
+                auto tradeTickets = CArtTradeTicket::FindAllTicketByArtTnxID(t.GetTxId());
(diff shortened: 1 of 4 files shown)


src/mnode/ticket-processor.cpp
```diff
@@ -1087,6 +1087,80 @@ bool CPastelTicketProcessor::ParseP2FMSTransaction(const CMutableTransaction& tx
     return true;
 }
 
+std::vector<std::string> CPastelTicketProcessor::ValidateOwnership(const std::string &_txid, const std::string &_pastelID)
```
akobrin1, line 1090: I think this function is pretty inefficient. As far as I understand the purpose was to check if the art specified by txid belongs to the specified pastel Id. I think instead of listing all existing art trade tickets (this potentially can be a huge list)  and filtering them making again several times full traverse of blockchain, you can just list all trade tickets related to that art txid,
then simply build a trade chain ordered by block height and all sell/buy txids in art trade tickets.
This chain should end on the guy with the specified pastelid.
@ayk please correct me if I wrong

akobrin1 (changes requested):

src/mnode/ticket-processor.cpp
devcareer0: I tried with the following in the "else" clause:
 ```

                //List trade tickets by txID and rearrange them by blockheight
                std::vector<CArtTradeTicket> tradeTickets = CArtTradeTicket::FindAllTicketByArtTnxID(_txid);
                std::sort(tradeTickets.begin(), tradeTickets.end(), [](CArtTradeTicket & one, CArtTradeTicket & two){return one.GetBlock() < two.GetBlock();});

                //Last element shall be the same pastelID to be verified.
                if(tradeTickets[tradeTickets.size()-1].pastelID.compare(_pastelID) == 0)
                {
                    //He shall be the owner indeed !
                    sRetVal[0] = _txid;
                    sRetVal[1] = tradeTickets[tradeTickets.size()-1].GetTxId();
                }
```
Is it sg. closer to the desired one ? (Tests results are failing BTW :) )

Comment (devcareer0) [reply src/mnode/ticket-processor.cpp:1090]: I'm running now tests from mock-networks and initiating cli commands to understand more the logic.
What i see (and currently dont yet fully understand is).
I initiate 2 commands:
- tickets list art all
- tickets list trade all

What i search for is a match between 1 **tx_id** in the **list art all** with 1 **art-txid** in **list trade all,** but there is no matching. So somewhat that tx_id is not present in trade ticket, at least not in that format so I do not know how to properly filter.. because this : `std::vector<CArtTradeTicket> tradeTickets = CArtTradeTicket::FindAllTicketByArtTnxID(_txid);`  will result nothing.

Here is the log of the 2 commands (list art all, list trade all) output:
[list_art_all_list_trade_all.txt](https://github.com/pastelnetwork/pastel/files/6722179/list_art_all_list_trade_all.txt)

So please lead me to a direction.

Comment (devcareer0) [reply src/mnode/ticket-processor.cpp:1090]: If i am not mistaken: from art-reg tx_id -> act_txid and this act_txid will become the art_txid in that trade ticket, am I right ? So somewhat i need an efficient way or query this, but currently just brainstorming.

Comment (devcareer0) [reply src/mnode/ticket-processor.cpp:1090]: Or.. Code snippet I copied above is OK, just i'm testing with wrong tx_id.. ?

Comment (devcareer0) [reply src/mnode/ticket-processor.cpp:1090]: If I dig deeper into the specification in freedcamp and trying to understand also - not just code it :) :
```
tickets tools validateownership... txid pastelid passphrase`

    Check if the pastelid from the request is present at the local key storage and the passphrase from the request is correct 
    Get the Art Registration ticket by txid from the request
    **If the author in the ticket (this is the Artist’s PastelID) matches pastelid from the request**
        return 
```

Okay, but how is guarantees ownership ? This only returns a 'guarantee' that the pastelid from the request is the author and not necessarily the owner.

devcareer0: Also:
I can originate an art this way:
art reg -> act -> trade in a 'straight way'.

But backwards I find a trade ticket as 'art_txid'. (So trade-ticket in a trade-ticket).

Currently I do not see any 'easy' method to make this ownership query 'quick' and efficient.

PR #45, opened 2021-06-26
I returned change to the sender wallet

ak-toktor:


vadymeng0:
Could you please provide below things in the description?
    - What is this task about, and related Freedcamp task.
    - Test result.
    - (If possible) some short description on how you implement it.
 Thank you!!!

src/wallet/rpcwallet.cpp
```diff
@@ -3981,14 +3981,14 @@ UniValue z_getoperationstatus_IMPL(const UniValue& params, bool fRemoveFinishedO
 #define CTXIN_SPEND_DUST_SIZE   148
 #define CTXOUT_REGULAR_SIZE     34
 
-UniValue z_sendmany(const UniValue& params, bool fHelp)
+UniValue z_sendmanyimpl(const UniValue& params, bool fHelp, const std::string& functionName, bool returnChangeToSenderAddr)
```
vadymeng0, line 3984: These two parameter has the same meaning. Should we deduct one?

src/wallet/rpcwallet.cpp
```diff
@@ -4249,6 +4245,23 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
         contextualTx.nVersion = 2; // Tx format should support vjoinsplits 
     }
 
+    // push sender address to the recipients to return change back to  the sender 
+    if(returnChangeToSenderAddr) {
+        int nMinDepth = 1;
+        if (fromTaddr) {
+            CAmount nBalance = getBalanceTaddr(fromaddress, nMinDepth, false);
```
vadymeng0, line 4252: In the case we return change to sender address, a new operation will still be added to the global queue? Is that mean another transaction will be created?

src/wallet/rpcwallet.cpp
```diff
@@ -5058,6 +5080,7 @@ static const CRPCCommand commands[] =
     { "wallet",             "z_gettotalbalance",        &z_gettotalbalance,        false },
     { "wallet",             "z_mergetoaddress",         &z_mergetoaddress,         false },
     { "wallet",             "z_sendmany",               &z_sendmany,               false },
+    { "wallet",             "z_sendmanywithchangetosender", &z_sendmanywithchangetosender, false },
```
vadymeng0, line 5083: This is **NOT** following the latest requirement that is suggested by @a-ok123 and @Dicklesworthstone 
We need to add _additional flag_ to `z_sendmany`, so the implementation must support:
    - `z_sendmany` (without additional flag): just behave old implementation. 
    - `z_sendmany -changetosendingaddress`: do the change utxo change back to the old address.
But you are creating a brand new command `z_sendmanywithchangetosender`.

vadymeng0:
This is reimplemented in: https://github.com/pastelnetwork/pastel/pull/51

vadymeng0:
I close this PR be cause it basically do not solve the problem.

vadymeng0 (changes requested):
Basically this change is adding more transaction to send the change back to the sender, but it **_do NOT remove the part that new address is generated and the change is sending to that new address_**.

a-ok123:
replaced by PR51

PR #49, opened 2021-07-02
Create additional option for command "list trade available" to allow filter by PastelID

vadymeng0:
## What is this PR about?
Create additional option for command "list trade available"
Old:

>   tickets trade - List ALL art trade tickets. Without filter parameter lists ALL Trade tickets.
>             Filter:
>               all       - list all Trade tickets (including non-confirmed). Default.
>               available - lists never sold Trade tickets (without Sell tickets).
>               sold      - lists only sold Trade tickets (with Sell tickets).

New:

>  tickets trade - List ALL art trade tickets. Without filter parameter lists ALL Trade tickets.
>             Filter:
>               all       - list all Trade tickets (including non-confirmed). Default.
>               available - lists never sold Trade tickets (without Sell tickets).
>               sold      - lists only sold Trade tickets (with Sell tickets).
>             Optional parameters:
>               <pastelID> - apply filter on trade ticket that belong to the correspond pastelID only

## How to test this? 
Added some test cases in the mn_tickets.py

## What is the test result?
Built successfully
Passed all below test suite on local machine:
cd qa/test-suite
./mn_main.py
./mn_payment.py
./mn_tickets.py
./mn_tickets_validation.py
./mn_messaging.py
./full_test_suite.py btest
./full_test_suite.py gtest
./full_test_suite.py sec-hard
./full_test_suite.py no-dot-so
./full_test_suite.py util-test
./full_test_suite.py secp256k1
./full_test_suite.py libsnark
./full_test_suite.py univalue
./full_test_suite.py rpc-common
./full_test_suite.py rpc-ext
Test case that is failing:
./mn_governance.py <-- Failed on master branch also.
Test case that took too long to complete, and failed some test cases while running:
./full_test_suite.py rpc-mn


a-ok123 (dismissed):

src/mnode/ticket-processor.cpp
```diff
@@ -685,39 +696,61 @@ std::string CPastelTicketProcessor::ListFilterSellTickets(const short filter) co
...
-        });
+        }, checkConfirmation);
 }
 
-// 1 - expired;    2 - sold
-std::string CPastelTicketProcessor::ListFilterBuyTickets(const short filter) const
+// 0 - all, 1 - expired;    2 - sold
+std::string CPastelTicketProcessor::ListFilterBuyTickets(const short filter, const std::string& pastelID) const
 {
+    bool checkConfirmation = true;
```
VitaliyVK777, line 705: probable just write:
const bool checkConfirmation{filter > 0};
and remove
"else
            checkConfirmation = false;"

src/mnode/ticket-processor.cpp
```diff
@@ -645,12 +645,23 @@ std::string CPastelTicketProcessor::ListFilterActTickets(const short filter) con
             return true;
         });
 }
-// 1 - available; 2 - unavailable; 3 - expired; 4 - sold
-std::string CPastelTicketProcessor::ListFilterSellTickets(const short filter) const
+// 0 - all, 1 - available; 2 - unavailable; 3 - expired; 4 - sold
+std::string CPastelTicketProcessor::ListFilterSellTickets(const short filter, const std::string& pastelID) const
 {
+    bool checkConfirmation = true;
```
VitaliyVK777, line 651: probable just write:
const bool checkConfirmation{filter > 0};
and remove
"else
checkConfirmation = false;"

src/mnode/ticket-processor.cpp
```diff
@@ -685,39 +696,61 @@ std::string CPastelTicketProcessor::ListFilterSellTickets(const short filter) co
...
-        });
+        }, checkConfirmation);
 }
-// 1 - available;      2 - sold
-std::string CPastelTicketProcessor::ListFilterTradeTickets(const short filter) const
+
+// 0 - all, 1 - available;      2 - sold
+std::string CPastelTicketProcessor::ListFilterTradeTickets(const short filter, const std::string& pastelID) const
 {
+    bool checkConfirmation = true;
```
VitaliyVK777, line 731: probable just write:
const bool checkConfirmation{filter > 0};
and remove
"else
checkConfirmation = false;"

src/mnode/mnode-rpc.cpp
```diff
@@ -2168,36 +2176,119 @@ As json rpc
             break;
 
         case RPC_CMD_LIST::sell:
+        {
+            std::string pastelID;
+
+            if (params.size() > 2 && params[2].get_str() != "all" && params[2].get_str() != "available" && params[2].get_str() != "unavailable"
```
VitaliyVK777, line 2182: we have 3 similar like blocks with difference in first condition with params[2] checking. probably write some one function?

VitaliyVK777 (changes requested):

src/mnode/ticket-processor.cpp
vadymeng0: that will change the logic here :)

src/mnode/mnode-rpc.cpp
vadymeng0: the comparing is different, and the inside-block code is just too short to create a new function

src/mnode/ticket-processor.cpp
VitaliyVK777: It will not change logic ) we have checkConfirmation false only for filter == 0. It will work correct if
    
    const bool checkConfirmation{filter > 0}; // do not check confirmation for ALL
    if (filter == 0 && pastelID.empty()) {
      return ListTickets<CArtSellTicket>(); // get all
    }
    return filterTickets<CArtSellTicket>(...
    
    It is not error, just less code

VitaliyVK777 (changes requested):

src/mnode/ticket-processor.cpp
vadymeng0: I see, will fix

src/mnode/mnode-rpc.cpp
VitaliyVK777: It is not bug, I just mean to place all this code in one function as it is repeated in 3 times:

template <typename F>
bool ParseParams(const UniValue& params, F f, int& minheight, std::string& pastelID) {
            if (f()) {
                if (params[2].get_str().find_first_not_of("0123456789") == std::string::npos) {
                    // This means min_height is input.
                    minheight = get_number(params[2]);
                } else {
                    // This means pastelID is input
                    pastelID = params[2].get_str();
                }
            } else if (params.size() > 2) {
                filter = params[2].get_str();
                if (params.size() > 3) {
                    if (params[3].get_str().find_first_not_of("0123456789") == std::string::npos) {
                        // This means min_height is input.
                        minheight = get_number(params[3]);
                    } else {
                        // This means pastelID is input
                        pastelID = params[3].get_str();
                    }
                }
                if (params.size() > 4) {
                    pastelID = params[3].get_str();
                    minheight = get_number(params[4]);
                }
            }
}
auto func1 = [params]() -> bool {
  return params.size() > 2 && params[2].get_str() != "all" && params[2].get_str() != "available" && params[2].get_str() != "unavailable" && params[2].get_str() != "expired" && params[2].get_str() != "sold"
}
ParseParams(params, func1, minheight, pastelID)

VitaliyVK777 (changes requested):

src/mnode/mnode-rpc.cpp
vadymeng0: Advantages:
     - Save some lines of code, because it repeats 3 times, so we save 2 blocks of repeated code.
Disadvantages:
     - Extendibility: Buy, Sell, Trade tickets contains different properties. Potentially these commands will be extended in the future to support filter by those properties. The "common" part is no longer common.
     - Giving another level of complexity when reading the code.
After considering, I don't think it's worth to implement that common function.

VitaliyVK777 (dismissed):

src/mnode/ticket-processor.cpp
```diff
@@ -544,7 +544,7 @@ template std::string CPastelTicketProcessor::ListTickets<CArtTradeTicket>() cons
 template std::string CPastelTicketProcessor::ListTickets<CArtRoyaltyTicket>() const;
 
 template <class _TicketType, typename F>
-std::string CPastelTicketProcessor::filterTickets(F f) const
+std::string CPastelTicketProcessor::filterTickets(F f, bool checkConfirmation) const
```
akobrin1, line 547: const bool bCheckConfirmation or fCheckConfirmation

akobrin1 (dismissed):
Looks good

src/mnode/ticket-processor.cpp
vadymeng0: fixed

a-ok123 (approved):

PR #56, opened 2021-07-12
40603916: Add legroast key generation and signing/verification to `pastelid` APIs

akobrin1:
**40603915: Modify pastelid APIs by adding an additional parameter - `algorithm`**

**Added support for LegRoast key generation/storage/signing/verificatiom in pasteld**

**Secure Container enhancements:**
  - added secure container prefix (to distinguish with old PastelID private key file)
  - added ability to save public items to container. 
  Items are saved in json serialized to msgpack binary format.
  They are always written before secure items, so that they can be read without secure items.
  -   saved LegRoast base58 encoded public key as a public item "pubkey_legroast".
  -   function to read public items only from secure container
  
  New secure container structure in json format:
  {
    "version":1,
    "public_items": [
        {
            "type":"item_type_name",
            "data": binary_t
        },
        {
            "type":"item_type_name",
            "data": binary_t
        }
    ]
  }
  {
    "version":1,
    "timestamp": int64_t,
    "encryption": "xchacha20-poly1305",
    "secure_items": [
        {
            "type":"secure_item_type_name",
            "nonce": binary_t,
            "data": binary_t
        },
        {
            "type":"secure_item_type_name",
            "nonce": binary_t,
            "data": binary_t
        }
    ]
  }
  
  - secure container binary format:

   PastelSecureContainer[public_items size][public_items hash][public_items msgpack][secure_items msgpack]
   PastelSecureContainer - secure container prefix;
   [public_items size] - size of the serialized public items msgpack; uint64_t in network byte order
   [public_items hash] - 256-bit hash of the public items (32 bytes)
   
**CPastelID class (pastelid/pastel_key.h|cpp):**
  -  CreateNewPastelKeys:
		Generates new PastelID (EdDSA448) and LegRoast public/private key pairs.
		PastelID is base58 encoded with prefix and checksum at the end.
		LegRoast public key is base58 encoded with prefix and checksum at the end.
		Creates new secure container to store all items associated with new PastelID.
		LegRoast public key is save as a public item and accessible without passphrase (for signing verification).
		PastelID private key (ed448) and LegRoast private key are saved as a secure items.
 -   Sign:
	    Sign text with the private key associated with PastelID.
		Added support for signing with LegRoast algorithm (used LegRoast Middle algorithm, public key size 4096, private key size 16).
		Added flag fBase64 to control signature output format (base64 or raw) - combined old Sign/Sign64 into one function.
		Sign first tries to read all required private keys for signing from the secure container file. If it does not find secure container prefix,
		it tries to read old PKCS8-encoded file with ed448 private key for backward compatibility.
		If algorithm to sign is LegRoast, but LegRoast private key is missing (old PKCS8 file), new LegRoast private/public key pair is generated and 
		PKCS8 file is replaced with the new secure container that stores ed448 private key, LegRoast private key, LegRoast public key.
-    Verify:
		Verify signature with the public key associated with PastelID.
		Added support for the new LegRoast algorithm.
		LegRoast public key is read from the secure container as a public item (does not need passphrase to decrypt).
		Added fBase64 flag to accept signatures in both base64 and raw formats - combined old Verify/Verify64 functions.
   GetStoredPastelIDs:
		Get PastelIDs stored locally in pastelkeys folder.
		Added bPastelIdOnly flag to retrieve only PastelIDs or PastelIDs with LegRoast public keys.
		Added validation for the invalid files in the pastelkeys folder.
		Read LegRoast public key (bPastelIdOnly=false) from secure container file as a public item (no passphrase needed).
	All strings/vectors holding passphrase/private keys are cleaned up in memory.

 **pastelid RPC API enhancements:**
        _pastelid newkey "passphrase"_
		    Generates now both PastelID (EdDSA448 private/public key pair) and LegRoast private/public key pair (CreateNewPastelKeys).
		    Both PastelID and base58-encoded LegRoast public key are returned in the result json object.
	_pastelid list:_
		    Lists all internally stored PastelID and keys. Returns also LegRoast public keys associated with the PastelID.
	_pastelid sign "text" "PastelID" "passphrase" ("algorithm"):_
		    Sign "text" with the internally stored private key associated with the PastelID (algorithm: ed448 or legroast).
	_pastelid verify "text" "signature" "PastelID" ("algorithm"):_
		    Verify "text"'s "signature" with the private key associated with the PastelID (algorithm: ed448 or legroast).
		
 **other code improvements:**
  -optimized base58 encode/decode functions
  -RPC_CMD_PARSER enhancement to support commands with hyphens (like sign-by-key). Enum defined in source code as sign__by__key.
  -vector_to_string/string_to_vector optimized functions

**Visual Studio 2019:**
   -changed pasteld/gtest VS projects to include new files

**python tests:**
   -mn_tickets.py: enhanced pastelid_tests to include:
       -checking legroast public key generation on "pastelid newkey" command
	   -signing/verification using new LegRoast algorithm
	   -negative tests for verification with LegRoast algorithm
gtest:
   -secure container tests for public items


vadymeng0:
LGTM!

vadymeng0 (approved):
LGTM!

src/pastelid/pastel_key.cpp
```diff
@@ -0,0 +1,314 @@
...
+#include "pastelid/secure_container.h"
+
+/**
+* Generate new PastelID (EdDSA448) and LegRoast public/private key pairs.
+* Create new secure container to store all items associated with PastelID.
+* 
+* \param passPhrase - secure passphrase that will be used to encrypt secure container.
+* \return pastelid_store_t map [encoded PastelID] -> [encoded LegRoast public key]
+*/
+pastelid_store_t CPastelID::CreateNewPastelKeys(SecureString&& passPhrase)
```
VitaliyVK777, line 18: use return std::tuple< string, string > instead of map

VitaliyVK777 (changes requested):

src/vector_types.h
```diff
@@ -0,0 +1,50 @@
...
+}
+
+inline void append_string_to_vector(const std::string &s, v_uint8 &v) noexcept
+{
+    const size_t nSize = v.size();
+    v.resize(nSize + s.size());
+    memcpy(v.data() + nSize, s.data(), s.size());
+}
+
+inline std::string vector_to_string(const v_uint8 &v)
```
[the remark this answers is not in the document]
akobrin1: v.resize can throw an exception, it's declared without noexcept... so we can't use noexcept here.

src/vector_types.h
```diff
@@ -0,0 +1,50 @@
...
+{
+    std::string s;
+    s.resize(v.size());
```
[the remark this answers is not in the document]
akobrin1: s.resize(0) is absolutely valid call. For performance reasons - checking every time "if (!v.empty()))" will take more time than just execute s.resize(0) for one single case. Moreover - chances that it will be called with empty vector are close to 0.

src/vector_types.h
```diff
@@ -0,0 +1,50 @@
...
+
+inline void string_to_vector(const std::string &s, v_uint8 &v) noexcept
+{
+    v.resize(s.size());
+    memcpy(v.data(), s.data(), s.size());
+}
+
+inline void append_string_to_vector(const std::string &s, v_uint8 &v) noexcept
+{
+    const size_t nSize = v.size();
```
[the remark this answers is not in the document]
akobrin1: same reasoning as above. 99% cases it won't be called with empty vector, and s.resize(0) - won't throw an exception.

src/vector_types.h
```diff
@@ -0,0 +1,50 @@
...
+{
+    v_uint8 v;
+    v.resize(s.size());
+    memcpy(v.data(), s.data(), s.size());
+    return v;
+}
+
+inline void string_to_vector(const std::string &s, v_uint8 &v) noexcept
+{
+    v.resize(s.size());
```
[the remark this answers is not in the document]
akobrin1: vector.resize(0) is also valid call. In any case -> if empty string s was passed to this function: we must resize vector to 0... so don't see any valid point to check that s is not empty.

src/vector_types.h
```diff
@@ -0,0 +1,50 @@
...
+ * Convert string to byte vector.
+ * This function is using memcpy and ~60 times faster than using v.assign(s.cbegin(), s.cend())
+ * 
+ * \param s - string to convert
+ * \return - byte vector
+ */
+inline v_uint8 string_to_vector(const std::string &s) noexcept
+{
+    v_uint8 v;
+    v.resize(s.size());
```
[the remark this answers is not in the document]
akobrin1: same

src/utilstrencodings.cpp
```diff
@@ -99,10 +100,12 @@ bool IsHex(const string& str)
     return (str.size() > 0) && (str.size()%2 == 0);
 }
 
-vector<unsigned char> ParseHex(const char* psz)
+v_uint8 ParseHex(const char* psz)
 {
     // convert hex dump to vector
-    vector<unsigned char> vch;
+    v_uint8 vch;
+    const size_t nLength = psz ? strlen(psz) : 0;
```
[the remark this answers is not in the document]
akobrin1: vch has to be returned in any case, if nLength is 0 or not.
So why we need to add any extra checks here?
The point is - if code works in a special case (like this when nLength=0) without any problems, then why to introduce any extra checks? To reduce performance?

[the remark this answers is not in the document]
akobrin1: I've corrected ParseHex here only to improve performance -> to make reservation for the vector. Without that it was simply reallocating memory for every new char.

src/pastelid/secure_container.h
```diff
@@ -59,74 +80,148 @@ inline SECURE_ITEM_TYPE GetSecureItemTypeByName(const std::string &sType)
...
+            memory_cleanse(nonce.data(), nonce.size());
+            memory_cleanse(data.data(), data.size());
+            pHandler = nullptr;
+        }
 
-   CSecureContainer() : 
-       m_nVersion(SECURE_CONTAINER_VERSION),
-        m_nTimestamp(-1)
-   {}
+        SECURE_ITEM_TYPE type;
```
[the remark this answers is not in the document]
akobrin1: ?? there are two constructors (one default, one - that initializes all fields), where type is initialized properly.

src/pastelid/secure_container.h
```diff
@@ -59,74 +80,148 @@ inline SECURE_ITEM_TYPE GetSecureItemTypeByName(const std::string &sType)
...
+        nlohmann::json::binary_t nonce; // public nonce used to encrypt the data
+        nlohmann::json::binary_t data;  // secure item data
+        ISecureDataHandler* pHandler;
```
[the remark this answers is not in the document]
akobrin1: same thing -> no need to do double initialization, if constructors are used - and they are initializing all class members.

src/pastelid/secure_container.h
```diff
@@ -59,74 +80,148 @@ inline SECURE_ITEM_TYPE GetSecureItemTypeByName(const std::string &sType)
...
+    {
+        _public_item_t() : 
+            type(PUBLIC_ITEM_TYPE::not_defined)
+        {}
+        _public_item_t(const PUBLIC_ITEM_TYPE atype, nlohmann::json::binary_t&& adata) : 
+            type(atype),
+            data(std::move(adata))
+        {}
+
+        PUBLIC_ITEM_TYPE type;
```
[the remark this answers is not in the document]
akobrin1: same reasoning -> unless class members not initialized in constructor -> we have to use c++11 class member initialization at  declaration feature.

src/pastelid/secure_container.h
```diff
@@ -59,74 +80,148 @@ inline SECURE_ITEM_TYPE GetSecureItemTypeByName(const std::string &sType)
...
+    std::string extract_secure_data_string(const SECURE_ITEM_TYPE type);
 
 private:
     static constexpr size_t PWKEY_BUFSUZE = crypto_box_SEEDBYTES;
 
     // header
-    uint16_t m_nVersion;  // container version
-    int64_t m_nTimestamp; // time stamp
+    uint16_t m_nVersion;                // container version
+    int64_t m_nTimestamp;               // time stamp
```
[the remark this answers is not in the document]
akobrin1: default constructor is doing all that

src/pastelid/secure_container.h
```diff
@@ -59,74 +80,148 @@ inline SECURE_ITEM_TYPE GetSecureItemTypeByName(const std::string &sType)
...
+    // Extract secure data from the container by type (returns string)
+    std::string extract_secure_data_string(const SECURE_ITEM_TYPE type);
 
 private:
     static constexpr size_t PWKEY_BUFSUZE = crypto_box_SEEDBYTES;
 
     // header
-    uint16_t m_nVersion;  // container version
-    int64_t m_nTimestamp; // time stamp
+    uint16_t m_nVersion;                // container version
```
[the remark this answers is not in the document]
akobrin1: default constructor is doing all that

src/rpc/rpc_consts.h
```diff
@@ -1,17 +1,18 @@
...
+// Copyright (c) 2021 Pastel Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+// rpc response object keys
+constexpr auto RPC_KEY_RESULT					= "result";
+constexpr auto RPC_KEY_ERROR_MESSAGE			= "errorMessage";
+constexpr auto RPC_KEY_STATUS					= "status";
+constexpr auto RPC_KEY_ALIAS					= "alias";
+constexpr auto RPC_KEY_TXID					    = "txid";
```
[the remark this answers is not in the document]
akobrin1: corrected offsets

src/pastelid/secure_container.h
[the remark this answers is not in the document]
VitaliyVK777: I just initialize type by default where class member is declared. It was not bug here.

src/pastelid/secure_container.cpp
```diff
@@ -102,107 +182,283 @@ void CSecureContainer::clear() noexcept
...
-
-    try
+    do
     {
-        ifstream fs(sFilePath, ios::in | ios::ate | ios::binary);
-        fs.exceptions(std::ifstream::failbit | std::ifstream::badbit);
-        vector<uint8_t> v;
-        v.resize(fs.tellg());
+        // get file size
+        const auto nFileSize = fs.tellg();
```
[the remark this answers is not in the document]
akobrin1: file is open with ios::ate flag, so cursor is moved to the end of the stream.
It is done intentionally to get file size first, then use it to make memory allocations.

src/pastelid/secure_container.cpp
```diff
@@ -102,107 +182,283 @@ void CSecureContainer::clear() noexcept
...
+        if (nFileSize < 0)
+            break;
+        nDataSize = static_cast<uint64_t>(nFileSize);
         fs.seekg(0);
+        // read prefix from the file and compare with SECURE_CONTAINER_PREFIX
+        constexpr auto nPrefixLength = std::char_traits<char>::length(SECURE_CONTAINER_PREFIX);
+        if (nDataSize < nPrefixLength)
+            break;
```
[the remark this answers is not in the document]
akobrin1: does not make big difference, will move

src/pastelid/secure_container.cpp
```diff
@@ -26,28 +66,48 @@ void CSecureContainer::add_item(const SECURE_ITEM_TYPE type, const nlohmann::jso
  * \param sPassphrase - passphrase in clear text to use for encryption
  * \return true if file was successfully written
  */
-bool CSecureContainer::write_to_file(const string& sFilePath, SecureString&& sPassphrase)
+bool CSecureContainer::write_to_file(const string& sFilePath, const SecureString& sPassphrase)
 {
     using json = nlohmann::ordered_json;
     bool bRet = false;
     do
```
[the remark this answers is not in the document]
akobrin1: ok

src/utilstrencodings.cpp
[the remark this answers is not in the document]
VitaliyVK777: I thought to add return to exclude moving to next code lines (reserve, while{})

VitaliyVK777 (changes requested):

src/pastelid/common.h
```diff
@@ -7,62 +7,73 @@
...
-                valb-=6;
+            while (valb >= 0)
+            {
+                out.push_back(BASE64[(val >> valb) & 0x3F]);
+                valb -= 6;
             }
         }
-        if (valb>-6) out.push_back("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/"[((val<<8)>>(valb+8))&0x3F]);
-        while (out.size()%4) out.push_back('=');
+        if (valb>-6)
```
VitaliyVK777, line 33: missing spaces

src/pastelid/common.h
```diff
@@ -7,62 +7,73 @@
...
-        while (out.size()%4) out.push_back('=');
+        if (valb>-6) 
+            out.push_back(BASE64[((val << 8) >> (valb + 8)) & 0x3F]);
+        while (out.size() % 4)
+            out.push_back('=');
         return out;
     }
 	
-	inline std::vector<unsigned char> Base64_Decode(const std::string &in)
+	inline v_uint8 Base64_Decode(const std::string &in) noexcept
```
VitaliyVK777, line 40: offset

VitaliyVK777: const std::string& in

src/pastelid/common.h
```diff
@@ -7,62 +7,73 @@
...
+#include "vector_types.h"
 
 namespace ed_crypto {
 
     static constexpr int OK = 1;
-	
-	inline std::string Base64_Encode(const unsigned char *in, size_t len)
+    static constexpr auto BASE64 = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
+
+	inline std::string Base64_Encode(const unsigned char *in, const size_t len) noexcept
```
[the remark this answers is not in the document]
VitaliyVK777: const unsigned char* in

src/pastelid/common.h
```diff
@@ -7,62 +7,73 @@
...
-	inline std::vector<unsigned char> Base64_Decode(const std::string &in)
+	inline v_uint8 Base64_Decode(const std::string &in) noexcept
     {
-        std::vector<unsigned char> out;
+        v_uint8 out;
+        out.reserve((in.size() / 4) * 3);
 
         std::vector<int> T(256,-1);
-        for (int i=0; i<64; i++) T["ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/"[i]] = i;
+        for (int i=0; i<64; i++)
```
VitaliyVK777, line 46: missing spaces

src/pastelid/common.h
```diff
@@ -7,62 +7,73 @@
...
+            T[BASE64[i]] = i;
 
         int val=0, valb=-8;
-        for (unsigned char c : in) {
-            if (T[c] == -1) break;
-            val = (val<<6) + T[c];
+        for (unsigned char c : in)
```
VitaliyVK777, line 50: const unsigned char c

src/pastelid/common.h
```diff
@@ -7,62 +7,73 @@
...
+        {
+            if (T[c] == -1)
+                break;
+            val = (val << 6) + T[c];
             valb += 6;
-            if (valb>=0) {
-                out.push_back( (unsigned char)((val>>valb)&0xFF));
+            if (valb >= 0)
+            {
+                out.push_back( (unsigned char)((val >> valb) & 0xFF));
```
VitaliyVK777, line 58: static_cast< unsigned char >

src/pastelid/common.h
```diff
@@ -7,62 +7,73 @@
...
-        std::vector<unsigned char> out;
+        v_uint8 out;
+        out.reserve((in.size() / 4) * 3);
 
         std::vector<int> T(256,-1);
-        for (int i=0; i<64; i++) T["ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/"[i]] = i;
+        for (int i=0; i<64; i++)
+            T[BASE64[i]] = i;
 
         int val=0, valb=-8;
```
VitaliyVK777, line 49: missing spaces

src/pastelid/common.h
```diff
@@ -7,62 +7,73 @@
...
+            if (T[c] == -1)
+                break;
+            val = (val << 6) + T[c];
             valb += 6;
-            if (valb>=0) {
-                out.push_back( (unsigned char)((val>>valb)&0xFF));
+            if (valb >= 0)
+            {
+                out.push_back( (unsigned char)((val >> valb) & 0xFF));
                 valb-=8;
```
VitaliyVK777, line 59: spaces

src/pastelid/common.h
```diff
@@ -7,62 +7,73 @@
...
             }
         }
         return out;
     }
 	
-	inline std::string Hex_Encode(const unsigned char *in, size_t len)
+	inline std::string Hex_Encode(const unsigned char *in, const size_t len)
```
VitaliyVK777, line 65: offset

VitaliyVK777: const unsigned char* in

src/pastelid/common.h
```diff
@@ -7,62 +7,73 @@
...
     static constexpr int OK = 1;
-	
-	inline std::string Base64_Encode(const unsigned char *in, size_t len)
+    static constexpr auto BASE64 = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
+
+	inline std::string Base64_Encode(const unsigned char *in, const size_t len) noexcept
     {
         std::string out;
+        out.reserve(static_cast<size_t>(ceil(len / 3) * 4));
         int val=0, valb=-6;
```
VitaliyVK777, line 21: missing spaces

src/pastelid/common.h
```diff
@@ -7,62 +7,73 @@
...
     {
         std::ostringstream hex_str_str;
-        for (int i = 0; i < len; i++)
+        for (size_t i = 0; i < len; i++)
             hex_str_str << std::setfill('0') << std::setw(2) << std::hex << (int) in[i];
         return hex_str_str.str();
     }
 	
-	inline std::vector<unsigned char> Hex_Decode(const std::string &in)
+	inline v_uint8 Hex_Decode(const std::string &in)
```
VitaliyVK777, line 73: offset

VitaliyVK777: const std::string& in

src/pastelid/common.h
```diff
@@ -7,62 +7,73 @@
...
     {
-        std::vector<unsigned char> out;
-        for (int i=0;i<in.length();i+=2)
+        v_uint8 out;
+        for (size_t i = 0; i < in.length(); i+=2)
```
VitaliyVK777, line 76: spaces

src/pastelid/common.h
```diff
@@ -117,43 +128,48 @@ namespace ed_crypto {
         using unique_buffer_ptr = std::unique_ptr<unsigned char, BufferDeleterFunctor>;
 
     public:
+        buffer(unsigned char *pbuf, const std::size_t len) :
```
VitaliyVK777, line 131: unsigned char* pbuf

src/pastelid/common.h
```diff
@@ -117,43 +128,48 @@ namespace ed_crypto {
...
+        std::size_t m_nLength;
     };
-    class crypto_exception : public std::exception {
+
+    class crypto_exception : public std::exception
+    {
         std::string message;
     public:
-        crypto_exception(std::string error, std::string details, std::string func_name) {
+        crypto_exception(const std::string &error, const std::string &details, const std::string &func_name)
```
VitaliyVK777, line 171: const std::string& error ...

src/pastelid/common.h
```diff
@@ -166,7 +182,8 @@ namespace ed_crypto {
             message = str_str.str();
         }
 
-        const char *what() const noexcept {
+        const char *what() const noexcept
```
VitaliyVK777, line 185: char*

src/mnode/mnode-rpc.cpp
```diff
@@ -40,7 +40,8 @@ UniValue formatMnsInfo(const std::vector<CMasternode>& topBlockMNs)
 
     int i = 0;
     KeyIO keyIO(Params());
-    for (auto &mn : topBlockMNs) {
+    for (const auto &mn : topBlockMNs)
```
VitaliyVK777, line 43: const auto& mn

src/mnode/mnode-rpc.cpp
```diff
@@ -595,7 +596,7 @@ UniValue masternode(const UniValue& params, bool fHelp)
         strKeyPass.reserve(100);
         strKeyPass = params[4].get_str().c_str();
         
-        std::string pastelID = CPastelID::CreateNewLocalKey(strKeyPass);
+        std::string pastelID = CPastelID::CreateNewLocalKey(std::move(strKeyPass));
```
VitaliyVK777, line 599: strKeyPass is used on the next line

src/mnode/mnode-rpc.cpp
```diff
@@ -1216,159 +1217,184 @@ UniValue governance(const UniValue& params, bool fHelp)
...
-  sign "text" "PastelID" "passphrase"        - Sign "text" with the internally stored private key associated with the PastelID.
-  sign_by_key "text" "key" "passphrase"      - Sign "text" with the private "key" (EdDSA448) as PKCS8 encrypted string in PEM format.
-  verify "text" "signature" "PastelID"       - Verify "text"'s "signature" with the PastelID.
+  newkey "passphrase"                                - Generate new PastelID, associated keys (EdDSA448) and LegRoast signing keys.
+                                                       Return PastelID and LegRoast signing public key base58-encoded.
+                                                       "passphrase" will be used to encrypt the key file.
+  importkey "key" <"passphrase">                     - Import private "key" (EdDSA448) as PKCS8 encrypted string in PEM format. Return PastelID base58-encoded
+                                                       "passphrase" (optional) to decrypt the key for the purpose of validating and returning PastelID.
+  											           NOTE: without "passphrase" key cannot be validated and if key is bad (not EdDSA448) call to "sign" will fail
```
VitaliyVK777, line 1246: NOTE offset

src/mnode/mnode-rpc.cpp
```diff
@@ -1216,159 +1217,184 @@ UniValue governance(const UniValue& params, bool fHelp)
...
         }
 
         return resultArray;
     }
+
+    // sign text with the internally stored private key associated with the PastelID (ed448 or legroast).
     if (PASTELID.IsCmd(RPC_CMD_PASTELID::sign))
     {
-        if (params.size() != 4)
+        if (params.size() < 4)
```
VitaliyVK777, line 1327: if (params.size() < 4 || params.size()>5)

src/mnode/mnode-rpc.cpp
```diff
@@ -1216,159 +1217,184 @@ UniValue governance(const UniValue& params, bool fHelp)
...
             throw JSONRPCError(RPC_INVALID_PARAMETER,
-				"pastelid sign \"text\" \"PastelID\" \"passphrase\"\n"
-				"Sign \"text\" with the internally stored private key associated with the PastelID."
-			);
+R"(pastelid sign "text" "PastelID" "passphrase" ("algorithm")
+Sign "text" with the internally stored private key associated with the PastelID.)");
```
VitaliyVK777, line 1330: probably write some information about "algorithm"?

src/mnode/mnode-rpc.cpp
```diff
@@ -1216,159 +1217,184 @@ UniValue governance(const UniValue& params, bool fHelp)
...
+        if (strKeyPass.empty())
             throw runtime_error(
-				"pastelid importkey <\"passphrase\">\n"
-				"passphrase for imported key cannot be empty!"
-			);
+R"(pastelid sign "text" "PastelID" <"passphrase">
+passphrase for the private key cannot be empty!)");
+
+        string sAlgorithm;
+        if (params.size() >= 5)
```
VitaliyVK777, line 1342: == 5

VitaliyVK777 (changes requested):

a-ok123 (dismissed):

src/mnode/mnode-rpc.cpp
akobrin1: well, let's say we have new UI version communicating with old pasteld daemon. In this case it theoretically can send more than 5 parameters. This old version will simply won't work in this case.

src/mnode/mnode-rpc.cpp
akobrin1: same reasoning, new UI communicating with old pasteld

src/pastelid/common.h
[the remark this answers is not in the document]
akobrin1: we're trying to check code errors here, not coding style preferences. C++ interprets "type& var" or "type &var" exactly the same, so I think we can safely write either way. It does not make an error.

src/pastelid/common.h
akobrin1: If you see one line change - it means that I "changed one line only" instead of rewriting the whole function.
In this case I just added string reservation to improve base64 encoding performance.

src/pastelid/pastel_key.cpp
```diff
@@ -0,0 +1,314 @@
...
+            continue;
+        sLegRoastKey.clear();
+        if (!bPastelIdOnly)
+        {
+            // read public items from secure container
+            CSecureContainer cont;
+            if (cont.read_public_from_file(p.path().string()))
+                cont.get_public_data(PUBLIC_ITEM_TYPE::pubkey_legroast, sLegRoastKey);
+        }
+        resultMap.emplace(std::move(sPastelID), sLegRoastKey);
```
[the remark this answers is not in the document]
akobrin1: yes, should be std::move here for sLegRoastKey

src/utilstrencodings.cpp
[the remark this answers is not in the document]
akobrin1: we need to avoid multiple return statements in the code if possible.
This is a best coding practice. Look at https://en.wikipedia.org/wiki/Return_statement (Multiple return statements section)...
"The most common problem in early exit is that cleanup or final statements are not executed – for example, allocated memory is not unallocated, or open files are not closed, causing leaks."
It's not that important in this case, but use it as a rule - good practice.

src/pastelid/common.h
akobrin1: why char*? 
this is how it is declared: 
https://en.cppreference.com/w/cpp/error/exception/what
what is missing here actually is override at the end

src/utilstrencodings.cpp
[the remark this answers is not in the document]
VitaliyVK777: ok. will follow

src/pastelid/common.h
VitaliyVK777: I mean just formatting
const char* what() const noexcept
as I see you are trying to place or replace "&" or "*" in the left side. and I'm follow the same in my coding.

src/pastelid/common.h
VitaliyVK777: ok. good then

src/pastelid/common.h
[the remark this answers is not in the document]
VitaliyVK777: ok. fine

src/pastelid/pastel_key.cpp
```diff
@@ -0,0 +1,314 @@
...
+    pastelid_store_t resultMap;
+    std::string sPastelID, sLegRoastKey;
+    v_uint8 vData;
+    for (const auto& p : fs::directory_iterator(pathPastelKeys))
+    {
+        sPastelID = p.path().filename().string();
+        // check if this file name is in fact encoded Pastel ID
+        if (!DecodePastelID(sPastelID, vData))
+            continue;
+        sLegRoastKey.clear();
```
VitaliyVK777, line 247: and this can be removed then
sLegRoastKey.clear();

VitaliyVK777 (changes requested):

VitaliyVK777 (changes requested):

src/pastelid/pastel_key.cpp
akobrin1: If I reuse any variable in a loop - I like to clean it up before each usage. You never know who will be changing this function and how this variable will be used in this loop later on. 
So cleanup seems to me like a good idea. 
Does not matter that in this case this string is moved to the pastelid storage. 
IMHO this contributes to safe and reliable code.

src/pastelid/secure_container.cpp
```diff
@@ -26,28 +66,48 @@ void CSecureContainer::add_item(const SECURE_ITEM_TYPE type, const nlohmann::jso
...
-        nlohmann::ordered_json j =
+        json jItems;
+        // generate json for the public items
+        json jPublic =
+        {
+                {"version", SECURE_CONTAINER_VERSION }
+        };
+        size_t nJsonPublicSize = 20; // used to estimate size of the json with public items
+        
+        for (auto& item: m_vPublicItems)
```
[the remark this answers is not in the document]
akobrin1: yes, good catch... I think I was moving before item.data vector to json, but then decided to keep it in a container class.

src/pastelid/pastel_key.cpp
VitaliyVK777: ok. good

src/pastelid/secure_container.cpp
```diff
@@ -102,107 +182,283 @@ void CSecureContainer::clear() noexcept
...
         v.clear();
+        // process public items
+        string sType;
+        for (auto& jItem : j.at("public_items"))
+        {
+            jItem["type"].get_to(sType);
+            public_item_t item;
+            item.type = GetPublicItemTypeByName(sType);
+            if (item.type == PUBLIC_ITEM_TYPE::not_defined)
+                throw runtime_error(strprintf("Public item type '%s' is not supported in the secure container", sType));
```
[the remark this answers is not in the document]
akobrin1: it is used before throw... how can I move it after?

VitaliyVK777:
your code & style coding is very good so almost impossible to find any error ;)

src/pastelid/secure_container.cpp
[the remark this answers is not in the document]
VitaliyVK777: something like?
const auto type =  GetPublicItemTypeByName(sType);
if (type == PUBLIC_ITEM_TYPE::not_defined)
  throw ...
public_item_t item{.type = type};

you can keep it like now too as no bug here

src/pastelid/secure_container.cpp
```diff
@@ -102,107 +182,283 @@ void CSecureContainer::clear() noexcept
...
+    clear();
 
-        CSodiumAutoBuf pw;
-        // allocate secure memory for the key, buffer is reused for all secure items
-        if (!pw.allocate(PWKEY_BUFSUZE))
-            throw runtime_error(strprintf("Failed to allocate memory (%zu bytes)", PWKEY_BUFSUZE));
+    ifstream fs(sFilePath, ios::in | ios::ate | ios::binary);
+    fs.exceptions(std::ifstream::failbit | std::ifstream::badbit);
+    uint64_t nDataSize = 0;
+    return read_public_items_ex(fs, nDataSize);
```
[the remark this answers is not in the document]
akobrin1: yes, seems like a good idea. 
caught system_error and generated error string here instead - if caller wants to get more info about an error reading container

src/pastelid/pastel_key.cpp
akobrin1: nope, map here is for fast search of pastel ids later in the code.
Also, value right now is just string representing LegRoast public key. But later on, it can be replaced with structure containing additional fields.

src/pastelid/common.h
akobrin1: old code, but yes - will change

src/pastelid/common.h
akobrin1: same

src/mnode/mnode-rpc.cpp
akobrin1: this is commented code

src/mnode/mnode-rpc.cpp
akobrin1: yes, but think about next UI version and even newer API.
We actually need here only first 5 parameters and don't care what goes next.
For backward compatibility with all future versions - it's good to use < > param count comparisons.

VitaliyVK777: agree

src/mnode/mnode-rpc.cpp
akobrin1: added

src/pastelid/secure_container.cpp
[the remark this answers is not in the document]
akobrin1: exactly... it's the same thing, but no extra temporary variable. Also item is not dynamically allocated here, so no leak.

src/mnode/mnode-rpc.cpp
akobrin1: Forgot about this one, will correct in the next checkin...

akobrin1:
Thank you! But this is mutual process, we learn from each other.

VitaliyVK777 (approved):

a-ok123 (approved):

PR #59, opened 2021-07-15
#41162566 - Security fixes in the NFT core logic

VitaliyVK777:
All 63 tests + new mn_expiration.py tests are successful


src/mnode/mnode-pastel.cpp
```diff
@@ -603,14 +603,77 @@ bool common_validation(const T& ticket, bool preReg, const std::string& strTnxId
...
-    std::string err;
     if (!pastelTicket->IsValid(false, ++depth)) {
-      throw std::runtime_error(strprintf("The %s ticket with this txid [%s] is invalid - %s", prevTicket, strTnxId, err));
+      throw std::runtime_error(strprintf("The %s ticket with this txid [%s] is invalid", prevTicket, strTnxId));
     }
     
     return true;
 }
 
+void trade_copy_validation(const std::string& nftTnxId, const std::vector<unsigned char>& signature) {
```
akobrin1, line 613: It would be nice to add comments in doxygen format (at least for some new functions) and describe parameters.

src/mnode/mnode-pastel.cpp
```diff
@@ -603,14 +603,77 @@ bool common_validation(const T& ticket, bool preReg, const std::string& strTnxId
...
+        "The trade ticket with txid [%s] referred by this trade ticket is invalid", nftTnxId));
+    }
+
+    totalCopies = 1;
+  } else {
+    throw std::runtime_error(strprintf(
+      "Unknown ticket with txid [%s] referred by this trade ticket is invalid", nftTnxId));
+  }
+
+  auto soldCopies{0};
```
akobrin1, line 662: I think it's better to declare type implicitly. Because it should have same type as totalCopies.

src/mnode/mnode-pastel.cpp
```diff
@@ -603,14 +603,77 @@ bool common_validation(const T& ticket, bool preReg, const std::string& strTnxId
...
+  const auto existingTradeTickets = CArtTradeTicket::FindAllTicketByArtTnxID(nftTnxId);
+  for (const auto& t: existingTradeTickets) {
+    if (t.signature != signature) {
+      ++soldCopies;
+    }
+  }
+
+  if (soldCopies >= totalCopies) {
+    throw std::runtime_error(strprintf(
+      "Invalid trade ticket - cannot exceed the total number of available copies [%d] with sold [%d] copies",
```
akobrin1, line 672: specifier for "unsigned short" is "%hu"

src/mnode/mnode-pastel.cpp
```diff
@@ -854,24 +921,31 @@ bool CArtSellTicket::IsValid(bool preReg, int depth) const
...
-    //1. check PastelID in this ticket matches PastelID in the referred ticket (Activation or Trade)
-    //2. Verify the art is not already sold
-    auto existingTradeTickets = CArtTradeTicket::FindAllTicketByArtTnxID(artTnxId);
-    auto soldCopies = existingTradeTickets.size();
-    auto existingSellTickets = CArtSellTicket::FindAllTicketByArtTnxID(artTnxId);
-    auto sellTicketsNumber = existingSellTickets.size();
-    auto totalCopies = 0;
+
+    // Check PastelID in this ticket matches PastelID in the referred ticket (Activation or Trade)
+    auto totalCopies{0};
```
akobrin1, line 926: should be size_t type.
I think by default c++ will make for 0 -> int type.
Then  for (soldCopies >= totalCopies)  comparison compiler will give a warning: signed and unsigned data types comparison

src/mnode/mnode-pastel.cpp
```diff
@@ -854,24 +921,31 @@ bool CArtSellTicket::IsValid(bool preReg, int depth) const
...
+    // Verify the NFT is not already sold or gifted
+    const auto verifyAvailableCopies = [this](const std::string& strTicket, unsigned short totalCopies) {
```
akobrin1, line 928: const size_t totalCopies, will then be compared with soldCopies which is vector::size() -> size_t

src/mnode/mnode-pastel.cpp
```diff
@@ -603,14 +603,77 @@ bool common_validation(const T& ticket, bool preReg, const std::string& strTnxId
...
+  unsigned short totalCopies{0};
+
+  uint256 txid;
+  txid.SetHex(nftTnxId);
+  auto nftTicket = CPastelTicketProcessor::GetTicket(txid);
+  if (!nftTicket) {
+    throw std::runtime_error(strprintf(
+      "The NFT ticket with txid [%s] referred by this trade ticket is not in the blockchain", nftTnxId));
+  }
+  if (nftTicket->ID() == TicketID::Activate) {
```
akobrin1, line 627: This code looks like copy&paste from CArtSellTicket::IsValid. 
Can we move it to a separate function and reuse in both places?
Duplicate code is so difficult to maintain, you change one place - but then forget to change all others.
Looks like the code in CArtSellTicket::IsValid is just getting some additional data from tickets.

qa/rpc-tests/mn_expiration.py
```diff
@@ -0,0 +1,577 @@
...
+
+
+    # ===============================================================================================================
+    def initialize(self):
+        print("== Initialize nodes ==")
+
+        self.nodes[self.mining_node_num].generate(10)
+
+        # generate pastelIDs & send coins
+        self.artist_pastelid1 = self.nodes[self.non_mn3].pastelid("newkey", "passphrase")["pastelid"]
```
akobrin1, line 313: I've created create_pastelid function in MasterNodeCommon class, so after we merge my LegRoast changes - this code can be updated

akobrin1: also moved "passphrase" to class variable

akobrin1 (changes requested):

src/mnode/mnode-pastel.cpp
VitaliyVK777: ok

src/mnode/mnode-pastel.cpp
VitaliyVK777: ok

src/mnode/mnode-pastel.cpp
VitaliyVK777: correct

src/mnode/mnode-pastel.cpp
VitaliyVK777: ok

qa/rpc-tests/mn_expiration.py
VitaliyVK777: ok

src/mnode/mnode-pastel.cpp
VitaliyVK777: in CArtSellTicket::IsValid we have also additional checking for PastelID and copies are checked only with additional condition if (preReg || !ticketFound). So a few small changes but functions looks similar.

VitaliyVK777: I can remove this new function 'trade_copy_validation' as we ignore old sell/buy tickets and it is working correct now with latest fixes in the tests => so this function 'trade_copy_validation' in the end of CArtTradeTicket::IsValid is never fails. We can keep this function 'trade_copy_validation' just to be sure that all is correct and still available copies to sell. what do you think?

src/mnode/mnode-pastel.cpp
VitaliyVK777: added

src/mnode/mnode-pastel.cpp
VitaliyVK777: fixed

src/mnode/mnode-pastel.cpp
VitaliyVK777: fixed

src/mnode/mnode-pastel.cpp
VitaliyVK777: fixed

src/mnode/mnode-pastel.cpp
VitaliyVK777: fixed

src/mnode/mnode-pastel.h
```diff
@@ -351,7 +351,7 @@ class CArtSellTicket : public CPastelTicket
     std::string ToJSON() const noexcept override;
     std::string ToStr() const noexcept override;
     bool IsValid(bool preReg, int depth) const override;
-    CAmount TicketPrice(const unsigned int nHeight) const noexcept override { return askedPrice/50; }
+    CAmount TicketPrice(const unsigned int nHeight) const noexcept override { return std::max(10u, askedPrice / 50); }
```
vadymeng0, line 354: I am not sure what is the correct calculation.
But what is the expected TicketPrice if askedPrice is, for example: 2545? I think it should be 51, instead of 50, right?

src/mnode/mnode-pastel.cpp
akobrin1: you can leave it as it is.
We can think - maybe we can write later on some helper templates to get ticket and its properties (for different types of tickets).

akobrin1 (approved):
"Additional fixes in release" does not tell much... maybe better to just use original freedcamp id + some descriptive name.
It could help to trace changes. Or create new freedcamp task.

VitaliyVK777:
> "Additional fixes in release" does not tell much... maybe better to just use original freedcamp id + some descriptive name.
> It could help to trace changes. Or create new freedcamp task.

I agree. I also thought how to name this task. I will create a new freedcamp task.

VitaliyVK777:
> "Additional fixes in release" does not tell much... maybe better to just use original freedcamp id + some descriptive name.
> It could help to trace changes. Or create new freedcamp task.

fixed

vadymeng0 (approved):

a-ok123 (approved):

PR #78, opened 2021-11-02
Red4sec - addressing Item #6

a-ok123:


src/key.h
```diff
@@ -183,7 +183,10 @@ struct CExtKey {
     template <typename Stream>
     void Unserialize(Stream& s)
     {
-        unsigned int len = ::ReadCompactSize(s);
+        unsigned int len = ::ReadCompactSize(s); //not using BIP32_EXTKEY_SIZE as max size here -> want to throw my own exception
```
akobrin1, line 186: ReadCompactSize returns uint64_t, for 64-bit it's the same as size_t.... and s.read expects size_t parameter, so:
const size_t len = ::ReadCompactSize(s);

src/key.h
```diff
@@ -183,7 +183,10 @@ struct CExtKey {
...
+        if (len != BIP32_EXTKEY_SIZE) {
```
akobrin1, line 187: BIP32_EXTKEY_SIZE we can also make size_t and convert to constexpr:
constexpr size_t BIP32_EXTKEY_SIZE = 74;

src/pubkey.h
```diff
@@ -229,10 +229,11 @@ struct CExtPubKey {
     template <typename Stream>
     void Unserialize(Stream& s)
     {
-        unsigned int len = ::ReadCompactSize(s);
-        unsigned char code[BIP32_EXTKEY_SIZE];
-        if (len != BIP32_EXTKEY_SIZE)
+        unsigned int len = ::ReadCompactSize(s); //not using BIP32_EXTKEY_SIZE as max size here -> want to throw my own exception
```
akobrin1, line 232: same thing here:
const size_t len = ...

src/serialize.h
```diff
@@ -272,7 +273,7 @@ void WriteCompactSize(Stream& os, const uint64_t nSize)
 }
 
 template<typename Stream>
-uint64_t ReadCompactSize(Stream& is)
+uint64_t ReadCompactSize(Stream& is, uint32_t max_size = MAX_DATA_SIZE)
```
akobrin1, line 276: it supports 64-bit type, so max_size should be also the same type: uint64_t.
uint64_t ReadCompactSize(Stream& is, const uint64_t max_size = MAX_DATA_SIZE);
it's ok that MAX_DATA_SIZE is uint32_t: uint32_t -> uint64_t conversion is OK.

akobrin1 (changes requested):

akobrin1 (approved):

PR #85, opened 2021-11-24
sync write into ticket db; leveldb's limit of open files = 1000; extr…

a-ok123:
…a error message


akobrin2 (approved):
BTW, we have to look also how we're setting file descriptors limit... This is the logic in init.cpp:
    nMaxConnections = GetArg("-maxconnections", DEFAULT_MAX_PEER_CONNECTIONS);
    nMaxConnections = std::max(std::min(nMaxConnections, (int)(FD_SETSIZE - nBind - MIN_CORE_FILEDESCRIPTORS)), 0);
    int nFD = RaiseFileDescriptorLimit(nMaxConnections + MIN_CORE_FILEDESCRIPTORS);
    if (nFD < MIN_CORE_FILEDESCRIPTORS)
        return InitError(_("Not enough file descriptors available."));
    if (nFD - MIN_CORE_FILEDESCRIPTORS < nMaxConnections)
        nMaxConnections = nFD - MIN_CORE_FILEDESCRIPTORS;
Because this defines how many files we can open simultaneously. There is no check for hard limit (RLIMIT_NOFILE), soft limit (RLIMIT_NOFILE) should be set no less than 1000.
So when if we raise the option in LevelDB - we have to raise this limit accordingly.
DEFAULT_MAX_PEER_CONNECTIONS=125
MIN_CORE_FILEDESCRIPTORS=0

akobrin1 (approved):
BTW, we have to look also how we're setting file descriptors limit... This is the logic in init.cpp:
nMaxConnections = GetArg("-maxconnections", DEFAULT_MAX_PEER_CONNECTIONS);
nMaxConnections = std::max(std::min(nMaxConnections, (int)(FD_SETSIZE - nBind - MIN_CORE_FILEDESCRIPTORS)), 0);
int nFD = RaiseFileDescriptorLimit(nMaxConnections + MIN_CORE_FILEDESCRIPTORS);
if (nFD < MIN_CORE_FILEDESCRIPTORS)
return InitError(_("Not enough file descriptors available."));
if (nFD - MIN_CORE_FILEDESCRIPTORS < nMaxConnections)
nMaxConnections = nFD - MIN_CORE_FILEDESCRIPTORS;
Because this defines how many files we can open simultaneously. There is no check for hard limit (RLIMIT_NOFILE), soft limit (RLIMIT_NOFILE) should be set no less than 1000.
So when we raise the option in LevelDB - we have to raise this limit accordingly.
DEFAULT_MAX_PEER_CONNECTIONS=125
MIN_CORE_FILEDESCRIPTORS=0

PR #88, opened 2021-12-15
PSL-119 - better error message

tanlemin91:
- need better error message about wrong  password

### DIFF
--- a/qa/rpc-tests/mn_tickets_username_change.py
+++ b/qa/rpc-tests/mn_tickets_username_change.py
@@ -160,7 +160,7 @@ class UserNameChangeTest(PastelTestFramework):
         assert_raises_rpc(rpc.RPC_MISC_ERROR, self.ERR_READ_PASTELID_FILE, 
             self.nodes[2].tickets, "register", "username", username2, self.n1_pastelid1, self.passphrase)
         # using invalid passphrase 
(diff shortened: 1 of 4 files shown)


akobrin2 (changes requested):
There are multiple reasons why secure container cannot be read:
1) no access to file or file does not exist
2) file is locked for read
3) invalid file format - any validation fails
The only case when it fails because of incorrect passphrase is when crypto_aead_xchacha20poly1305_ietf_decrypt function fails and returns "Failed to decrypt secure item data' message. Just set some bool flag here and handle it in catch (const std::exception &ex). Or, even better, define custom exception from runtime_exception (let's say secure_container_exception with specific code & message), then add one more catch for that new exception.

tanlemin91:
> There are multiple reasons why secure container cannot be read:
> 
> 1. no access to file or file does not exist
> 2. file is locked for read
> 3. invalid file format - any validation fails
>    The only case when it fails because of incorrect passphrase is when crypto_aead_xchacha20poly1305_ietf_decrypt function fails and returns "Failed to decrypt secure item data' message. Just set some bool flag here and handle it in catch (const std::exception &ex). Or, even better, define custom exception from runtime_exception (let's say secure_container_exception with specific code & message), then add one more catch for that new exception.

a-ok123 (approved):

akobrin1 (approved):
looks good

a-ok123 (approved):

PR #91, opened 2021-12-19
get height from mempool lookup

tanlemin91:


src/rest.cpp
```diff
@@ -353,7 +353,8 @@ static bool rest_tx(HTTPRequest* req, const std::string& strURIPart)
 
     CTransaction tx;
     uint256 hashBlock = uint256();
-    if (!GetTransaction(hash, tx, Params().GetConsensus(), hashBlock, true))
+    uint32_t height;
+    if (!GetTransaction(hash, tx, Params().GetConsensus(), hashBlock, &height, true))
```
akobrin1, line 357: don't need height here -> nullptr
as you see, fAllowSlow parameter is changing pretty often, so it would make sense to move pnBlockHeight parameter after fAllowSlow.
This way - you don't need to change so many places, just set default value in main.h file:
   bool fAllowSlow = false, uint32_t *pnBlockHeight = nullptr, ....

src/rpc/rawtransaction.cpp
```diff
@@ -298,7 +298,8 @@ Result (if verbose > 0):
 
     CTransaction tx;
     uint256 hashBlock;
-    if (!GetTransaction(hash, tx, Params().GetConsensus(), hashBlock, true, blockindex))
+    uint32_t height;
+    if (!GetTransaction(hash, tx, Params().GetConsensus(), hashBlock, &height, true, blockindex))
```
akobrin1, line 302: pass nullptr

src/rpc/rawtransaction.cpp
```diff
@@ -384,7 +385,8 @@ UniValue gettxoutproof(const UniValue& params, bool fHelp)
     if (!pblockindex)
     {
         CTransaction tx;
-        if (!GetTransaction(oneTxid, tx, consensusParams, hashBlock, false) || hashBlock.IsNull())
+        uint32_t height;
+        if (!GetTransaction(oneTxid, tx, consensusParams, hashBlock, &height, false) || hashBlock.IsNull())
```
akobrin1, line 389: won't change if you move pnBlockHeight parameter after fAllowSlow

src/txmempool.h
```diff
@@ -215,7 +215,7 @@ class CTxMemPool
     }
 
     // Lookup for the transaction with the specific hash (txid).
-    virtual bool lookup(const uint256 &txid, CTransaction& tx) const;
+    virtual bool lookup(const uint256 &txid, CTransaction& tx, uint32_t &nBlockHeight) const;
```
akobrin1, line 218: I think you can even change uint32_t &nBlockHeight parameter to uint32_t *pnBlockHeight.
and again: if (pnBlockHeight) -> you have to make sure it is assigned in ALL paths inside CMemPool::lookup.
If you make that parameter optional -> many places you've changed will also go away.

akobrin1 (changes requested):
Please check my comments.
Also, you have to add google tests for changes like this (for mempool.lookup, GetTransaction) to cover all possible cases.

src/mnode/mnode-masternode.cpp
```diff
@@ -279,7 +279,8 @@ bool CMasternode::IsInputAssociatedWithPubkey()
 
     CTransaction tx;
     uint256 hash;
-    if (GetTransaction(vin.prevout.hash, tx, m_chainparams.GetConsensus(), hash, true))
+    uint32_t height;
+    if (GetTransaction(vin.prevout.hash, tx, m_chainparams.GetConsensus(), hash, &height, true))
```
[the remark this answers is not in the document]
tanlemin91: fixed. Thanks

src/mnode/mnode-masternode.cpp
```diff
@@ -686,7 +687,8 @@ bool CMasternodeBroadcast::CheckOutpoint(int& nDos)
     // should be at least not earlier than block when 1000 PASTEL tx got nMasternodeMinimumConfirmations
     uint256 hashBlock = uint256();
     CTransaction tx2;
-    GetTransaction(vin.prevout.hash, tx2, m_chainparams.GetConsensus(), hashBlock, true);
+    uint32_t height;
+    GetTransaction(vin.prevout.hash, tx2, m_chainparams.GetConsensus(), hashBlock, &height, true);
```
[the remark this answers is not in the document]
tanlemin91: fixed, Thanks

src/mnode/ticket-processor.cpp
```diff
@@ -483,6 +485,8 @@ unique_ptr<CPastelTicket> CPastelTicketProcessor::GetTicket(const uint256 &txid)
         int ticketBlockHeight = -1;
```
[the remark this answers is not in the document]
tanlemin91: fixed. Thanks

src/mnode/ticket-processor.cpp
```diff
@@ -464,7 +464,9 @@ unique_ptr<CPastelTicket> CPastelTicketProcessor::GetTicket(const uint256 &txid)
 {
     CTransaction tx;
     uint256 hashBlock;
-    if (!GetTransaction(txid, tx, Params().GetConsensus(), hashBlock, true))
+    uint32_t ticketHeight;
```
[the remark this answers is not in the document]
tanlemin91: Fixed. Thanks

src/rpc/rawtransaction.cpp
tanlemin91: fixed.

src/rpc/rawtransaction.cpp
tanlemin91: fixed, thanks

src/main.cpp
```diff
@@ -1510,16 +1510,20 @@ bool AcceptToMemoryPool(
...
     if (!blockIndex)
     {
-        if (mempool.lookup(hash, txOut))
+        if (mempool.lookup(hash, txOut, blkHeight))
+        {
+            *pnBlockHeight = blkHeight;
             return true;
+        }
 
         if (fTxIndex)
```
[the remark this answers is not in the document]
tanlemin91: fixed. Thanks for suggestion

src/main.cpp
```diff
@@ -1510,16 +1510,20 @@ bool AcceptToMemoryPool(
...
+    uint32_t blkHeight;
 
     LOCK(cs_main);
 
     if (!blockIndex)
     {
-        if (mempool.lookup(hash, txOut))
+        if (mempool.lookup(hash, txOut, blkHeight))
+        {
+            *pnBlockHeight = blkHeight;
```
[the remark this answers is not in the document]
tanlemin91: the gtest for mempool.lookup, GetTransaction, I still have no idea how to implement it. Can I own this until I complete all gtest tasks?

src/main.cpp
```diff
@@ -1510,16 +1510,18 @@ bool AcceptToMemoryPool(
  * Return transaction in txOut, and if it was found inside a block, its hash is placed in hashBlock.
  * If blockIndex is provided, the transaction is fetched from the corresponding block.
  */
-bool GetTransaction(const uint256 &hash, CTransaction &txOut, const Consensus::Params& consensusParams, uint256 &hashBlock, bool fAllowSlow, CBlockIndex* blockIndex)
+bool GetTransaction(const uint256 &hash, CTransaction &txOut, const Consensus::Params& consensusParams, uint256 &hashBlock, bool fAllowSlow, uint32_t *pnBlockHeight, CBlockIndex* blockIndex)
```
akobrin1, line 1513: there are so many paths in that function with return statement.
You can use the following technique in this case:
bool bRet = false;
uint32_t nBlockHeight = numeric_limits<uint32_t>::max();
do
{
      code_block1 that sets(or doesn't) nBlockHeight, 
            break instead of return
      code_block2 that sets(or doesn't) nBlockHeight, 
            break instead of return
     ..................
} while (false);
if (pnBlockHeight)
   *pnBlockHeight = nBlockHeight;
return bRet;

that way you have full control over setting pnBlockHeight. Also it's much easier to debug the function which is written like this.

akobrin1 (changes requested):

a-ok123 (approved):

akobrin1 (approved):

PR #93, opened 2021-12-22
Psl 170 conver univalue bench32 bip32 scriptnum

tanlemin91:
- convert univalue_tests.cpp, bech32_tests.cpp,
-  bip32_tests.cpp, scriptnum_tests.cpp to pastel_gtest


src/Makefile.gtest.include
```diff
@@ -69,7 +69,10 @@ pastel_gtest_SOURCES +=\
 	gtest/test_txid.cpp\
 	gtest/test_upgrades.cpp\
 	gtest/test_validation.cpp\
-	gtest/test_zip32.cpp
+	gtest/test_zip32.cpp\
```
akobrin1, line 72: try to order files by name - this way it's much easier to find smth

src/gtest/test_bip32.cpp
```diff
@@ -1,32 +1,38 @@
...
 
 #include "key.h"
 #include "key_io.h"
 #include "uint256.h"
 #include "util.h"
 #include "utilstrencodings.h"
-#include "test/test_bitcoin.h"
+#include "vector_types.h"
 
+#include <gtest/gtest.h>
```
akobrin1, line 14: We can use the following logic when adding includes:
make 3 separate groups of includes:
    - standard c++ includes
    - 3rdparty includes, like gtest.h. json.hpp, etc... all depends should go here
    - src includes:
         1.  more common includes (utils, encoding, etc) should go first, 
         2. then Pastel includes like src/pastelid, src/mnode/...
         3. any test-specific includes

src/gtest/test_bip32.cpp
```diff
@@ -36,6 +42,22 @@ struct TestVector {
     }
 };
 
+class TestBip32 : public Test
+{
+public:
+    void SetUp() override
```
akobrin1, line 48: you have to set chain params only once... 
so change this to:
static void SetUpTestSuite()
{
   ....
}
This will be executed  before all tests in this fixture once.

src/gtest/test_bip32.cpp
```diff
@@ -36,6 +42,22 @@ struct TestVector {
...
+
+    void TearDown() override
+    {
+    }
+
+
+};
+
+
 TestVector test1 =
```
akobrin1, line 61: This is classic candidate for parameterized test.
two tests with only one parameter (TestVector).
class PTestBip32 : public TestWithParam<TestVector>
{}
...
TEST_P instead of TEST_F
{
     const auto &test = GetParam();
    ..... whatever is in RunTest
}

INSTANTIATE_TEST_SUITE_P(bip32, PTestBip32, 
   Values(
        TestVector1,
        TestVector2
    ));

src/gtest/test_univalue.cpp
```diff
@@ -0,0 +1,316 @@
...
+TEST(test_univalue, univalue_readwrite)
+{
+    UniValue v;
+    EXPECT_TRUE(v.read(json1));
+
+    string strJson1(json1);
+    EXPECT_TRUE(v.read(strJson1));
+
+    EXPECT_TRUE(v.isArray());
+    EXPECT_EQ(v.size(), 2);
```
akobrin1, line 297: this usually gives a compilation warning.. if unsigned compared in EXPECT_EQ with signed.
add 'u' suffix in all places like this:
     EXPECT_EQ(v.size(), 2u)

akobrin1 (changes requested):
try to implement parameterized test PTestBip32
and merge Makefiles with a master branch

a-ok123 (approved):

src/Makefile.gtest.include
tanlemin91: fixed.Thanks

src/gtest/test_bip32.cpp
tanlemin91: Fixed. Thanks

src/gtest/test_bip32.cpp
tanlemin91: Fixed. Thanks

src/gtest/test_univalue.cpp
tanlemin91: Fixed. Thanks

akobrin1 (approved):
looks good

PR #103, opened 2022-01-12
Psl 95 filter out "NEW_START_REQUIRED" MNs that are dead for more than 1 week

tanlemin91:


src/mnode/mnode-controller.h
```diff
@@ -76,7 +76,7 @@ class CMasterNodeController
 
     int MasternodeCheckSeconds, MasternodeMinMNBSeconds, MasternodeMinMNPSeconds, MasternodeExpirationSeconds, MasternodeWatchdogMaxSeconds, MasternodeNewStartRequiredSeconds;
     int MasternodePOSEBanMaxScore;
-
+    int MasternodeWeekBySeconds;
```
akobrin1, line 79: This timeout variable meaning is MN expiration time for new_start_required status. I think MasterNodeWeekBySeconds does not reflect that meaning. You can call it smth like MNStartRequiredExpirationTime.
- make comments always when you add new member variable

src/mnode/rpc/masternode.cpp
```diff
@@ -117,6 +117,10 @@ UniValue masternodelist(const UniValue& params, bool fHelp)
         std::map<COutPoint, CMasternode> mapMasternodes = masterNodeCtrl.masternodeManager.GetFullMasternodeMap();
         for (auto& mnpair : mapMasternodes) {
             CMasternode mn = mnpair.second;
+            if( mn.IsNewStartRequired() && ! mn.IsPingedWithin(masterNodeCtrl.MasternodeWeekBySeconds) )
```
akobrin1, line 120: This will only skip MNs with the NewStartRequired expired status.
But the task was saying about new parameter to masternode & masternodelist RPC API to force show all MNs even with this expired status.

akobrin1 (changes requested):
- add rpc parameter to show all MNs (all commands that filter out MNs with this expired status)
- rebase this branch on master

src/mnode/rpc/masternode.cpp
```diff
@@ -55,11 +55,20 @@ UniValue masternodelist(const UniValue& params, bool fHelp)
 {
     std::string strMode = "status";
     std::string strFilter;
+    std::string strExtra = "";
```
akobrin1, line 58: You don't need to assign "" to strExtra, string default constructor sets it to empty.

src/mnode/rpc/masternode.cpp
```diff
@@ -117,6 +127,10 @@ UniValue masternodelist(const UniValue& params, bool fHelp)
         std::map<COutPoint, CMasternode> mapMasternodes = masterNodeCtrl.masternodeManager.GetFullMasternodeMap();
         for (auto& mnpair : mapMasternodes) {
             CMasternode mn = mnpair.second;
+            if( mn.IsNewStartRequired() && ! mn.IsPingedWithin(masterNodeCtrl.MNStartRequiredExpirationTime) && strExtra != "allnode")
```
akobrin1, line 130: you can calculate visibility filter once outside the loop and use it here:
const bool bShowAllNodes = strExtra == "allnode";

src/mnode/rpc/masternode.cpp
```diff
@@ -55,11 +55,20 @@ UniValue masternodelist(const UniValue& params, bool fHelp)
...
+    if (params.size() == 3)
+    {
+        strExtra = params[2].get_str();
+    }
+    else if (params.size() > 3)
+    {
+        throw JSONRPCError(RPC_INVALID_PARAMETER, "Too many parameters");
+    }
 
     if (fHelp || (strMode != "activeseconds" && strMode != "addr" && strMode != "full" && strMode != "info" &&
```
akobrin1, line 73: can you change this to RPC_CMD_PARSER? (look at any mnode/rpc/ticket-*.cpp files for example)

src/mnode/rpc/masternode.cpp
```diff
@@ -69,9 +78,10 @@ UniValue masternodelist(const UniValue& params, bool fHelp)
             "masternodelist ( \"mode\" \"filter\" )\n"
```
akobrin1, line 78: if (fHelp || !PASTELID.IsCmdSupported())
       throw runtime_error(
... string -> R-string.

akobrin1 (changes requested):

akobrin1 (approved):
thank you!

PR #105, opened 2022-01-20
Psl 207 convert pastel gtest

tanlemin91:


src/gtest/pastel_gtest_main.cpp
```diff
@@ -26,6 +28,7 @@ struct ECCryptoClosure
 
 ECCryptoClosure instance_of_eccryptoclosure;
 CPastelTest_Environment *gl_pPastelTestEnv = nullptr;
+fs::path pathTemp;
```
akobrin1, line 31: don't use global variable here, instead you can use the one that I have already in CPastelTest_Environment class:
m_TempDataDir.
This is used in InitializeChainTest/FinalizeChainTest and cleaned up in FinalizeChainTest via ClearTempDataDir().
I think you were moving this pathTemp from test_bitcoin.cpp, right?

src/gtest/test_rpc.cpp
```diff
@@ -1,33 +1,450 @@
...
-#include <primitives/block.h>
 #include <rpc/server.h>
-#include <streams.h>
+#include <rpc/client.h>
+#include <rpc/register.h>
+#include <key_io.h>
+#include <netbase.h>
+#include <main.h>
 #include <utilstrencodings.h>
+#include <univalue.h>
```
akobrin1, line 15: univalue.h, 3rdparty - can go up to the same place a gtest.h

src/gtest/test_rpc.cpp
```diff
@@ -1,33 +1,450 @@
...
+}
+
+class PTestRpc1 : public TestWithParam<tuple<string, CAmount, bool>>
+{};
+
+TEST_P(PTestRpc1, rpc_parse_monetary_values)
+{
+    const auto value = get<0>(GetParam());
+    const auto expectedValue = get<1>(GetParam());
+    const auto thow = get<2>(GetParam());
```
akobrin1, line 229: throw?

src/gtest/test_rpc_wallet.cpp
```diff
@@ -1,1258 +1,1274 @@
...
+#include <test_rpc.h>
+#include <pastel_gtest_main.h>
+
+using namespace std;
+using namespace testing;
+
+extern UniValue createArgs(int nRequired, const char* address1 = nullptr, const char* address2 = nullptr);
+extern UniValue CallRPC(string args);
+
+extern CWallet* pwalletMain;
```
akobrin1: this extern is declared already in init.h, we don't need it here... don't know why it was there in rpc_wallet_tests.cpp

src/gtest/test_rpc_wallet.cpp
```diff
@@ -1,1258 +1,1274 @@
...
+ * Tip: Create mock operations by subclassing AsyncRPCOperation.
+ */
+
+class MockSleepOperation : public AsyncRPCOperation {
+public:
+    chrono::milliseconds naptime;
+    MockSleepOperation(int t=1000) {
+        this->naptime = chrono::milliseconds(t);
+    }
+    virtual ~MockSleepOperation() {
```
akobrin1: remove it - default virtual destructor is declared in the parent class AsyncRPCOperation (BTW with noexcept)

src/gtest/test_rpc_wallet.cpp
```diff
@@ -1,1258 +1,1274 @@
...
+    }
+    virtual void main() {
```
akobrin1: this one better to change to:
void main() override

src/gtest/test_rpc_wallet.cpp
```diff
@@ -1,1258 +1,1274 @@
...
+    EXPECT_EQ(opids.count(id1) , 1U);
+    EXPECT_EQ(opids.count(id2) , 1U);
+    EXPECT_EQ(opids.count(op3->getId()), 1U);
+    q->finishAndWait();
+}
+
+// The CountOperation will increment this global
+atomic<int64_t> gCounter(0);
+
+class CountOperation : public AsyncRPCOperation {
```
akobrin1: same changes as for the MockSleepOperation class

src/univalue/lib/univalue_read.cpp
```diff
@@ -5,9 +5,10 @@
 #include <string.h>
 #include <vector>
 #include <stdio.h>
-#include "univalue.h"
+#include <univalue.h>
```
akobrin1, line 8: it's better to rollback changes in univalue files.
This is and external github project and unless we fix a bug - It's better to keep the original code here because this way it'll be much easier to move new changes they make

akobrin1 (changes requested):
few small changes

akobrin1 (changes requested):
find out why rpc common tests fail...

src/gtest/pastel_gtest_main.cpp
tanlemin91: done thanks

src/gtest/test_rpc.cpp
tanlemin91: done thanks

src/gtest/pastel_gtest_main.cpp
tanlemin91: previously, there is no InitializeChainTest/FinalizeChain in the pastel_gtest_main. Then I tried to setup one. Later you have done in the recent PR. Then I tried to merge master with your changes and still not removing all my changes in pastel_gtest_main. Fix Thanks

src/gtest/test_rpc_wallet.cpp
tanlemin91: thanks

src/gtest/test_rpc_wallet.cpp
tanlemin91: thanks

src/univalue/lib/univalue_read.cpp
tanlemin91: Don't know it is the third party. Fixed thanks

tanlemin91:
I tested 10 times, and could not see the issue. Do you encounter the issue before?

akobrin1 (approved):
looks like mempool sync is missing before generate somewhere in rest.py, will fix it in a separate story

PR #110, opened 2022-02-03
Psl 196 convert pastel gtest

tanlemin91:


akobrin1 (changes requested):
Please rebase and master and fix conflicts.
Also can you rename gtest/test_utils.cpp|h -> gtest/pastel_gtest_utils.cpp|h and correct all references in tests and projects. Let's keep all gtest helpers with different names (other than test_*).

akobrin1 (approved):

PR #122, opened 2022-04-13
masternodelist "extra" command missing filter functionality

TimothyBanks:
Added filters on the status and outpoint when executing the "extra' command on the masternodelist function.

### DIFF
--- a/src/mnode/rpc/masternode.cpp
+++ b/src/mnode/rpc/masternode.cpp
@@ -115,7 +115,7 @@ Examples:
         masterNodeCtrl.masternodeManager.UpdateLastPaid(pindex);
     }
 
-    auto is_filtered = [&](const auto&... filters) -> bool {
+    const auto is_filtered = [&](const auto&... filters) -> bool {
         return !strFilter.empty() && ((filters.find(strFilter) == std::string::npos) && ...);
     };
 
@@ -131,7 +131,7 @@ Examples:
             string strOutpoint = mnpair.second.vin.prevout.ToStringShort();
             if (is_filtered(strOutpoint))
                 continue;
-            obj.pushKV(strOutpoint, mnpair.first);
+            obj.pushKV(std::move(strOutpoint), mnpair.first);
         }
     } else {
         const auto mapMasternodes = masterNodeCtrl.masternodeManager.GetFullMasternodeMap();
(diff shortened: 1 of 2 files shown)


a-ok123 (dismissed):

src/mnode/rpc/masternode.cpp
```diff
@@ -115,6 +115,10 @@ Available modes:
         masterNodeCtrl.masternodeManager.UpdateLastPaid(pindex);
     }
 
+    auto is_filtered = [&](const auto&... filters) -> bool {
```
akobrin1, line 118: As I understand, the meaning of this lambda is the opposite: is_not_filtered - i.e. to check that strFilter is not in filters. If strFilter is defined - only items that pass this filter (that have substring) will be added to the output.

akobrin1: also can make this lambda const

src/mnode/rpc/masternode.cpp
```diff
@@ -146,16 +150,15 @@ Available modes:
             {
                 case RPC_CMD_MNLIST::activeseconds:
                 {
-                    if (!strFilter.empty() && strOutpoint.find(strFilter) == string::npos)
+                    if (is_filtered(strOutpoint))
                         continue;
                     obj.pushKV(strOutpoint, (int64_t)(mn.lastPing.sigTime - mn.sigTime));
```
akobrin1, line 155: BTW, can use obj.pushKV(move(strOutpoint), ....) in all cases

akobrin1 (changes requested):

src/mnode/rpc/masternode.cpp
TimothyBanks: The context that the lambda is being used is asking the question if the current item is filtered and if so, continuing to the next item.  I could change the conditionals to be something like `if (!is_not_filter(...)) continue;` but that's a double negative.

Sure, I can make it const.

src/mnode/rpc/masternode.cpp
TimothyBanks: Ok.  That seems outside the scope of this issue, but I can make that change.

src/mnode/rpc/masternode.cpp
TimothyBanks: I think the parameter name is what causes confusion.  `strFilter` is really a search term not a filter term.  We are using that variable to find the rows we want as opposed to the ros we don't want.

TimothyBanks:
Updated.

akobrin1 (changes requested):
Please rebase this branch on master. It may have some conflicts

TimothyBanks:
Do a rebase commit or a merge commit?

a-ok123:
We merge them

PR #123, opened 2022-04-13
Updated CMasternodeGovernance::VoteForTicket and CMasternodeGovernance::AddTicket pre-conditions

TimothyBanks:
### A. Improving governance voting logic

**Currently any active masternode can register governance payment ticket (CMasternodeGovernance::AddTicket), we need to add two more checks:**

check that MN is indeed enabled, by calling - CMasternode::IsEnabled() and check that it is “old enough”, this can be checked by comparing masternode’s broadcast message time with the current timethere is already method for that CMasternode::IsBroadcastedWithinold enough is 30 days (17280 blocks)

Same as above for Voting for ticket - CMasternodeGovernance::VoteForTicket

**B. Improving winner decision**

Governance ticket winners is decided if ticket gets 51% of Yes votes and 10% of masternode voted for it (CGovernanceTicket::IsWinner)

Currently 10% is calculated from the number of ALL masternodes in the network, this has to be changed to use only number of Enabled MNs

use CMasternodeMan::CountEnabled

### DIFF
--- a/src/mnode/mnode-governance.cpp
+++ b/src/mnode/mnode-governance.cpp
@@ -125,23 +125,23 @@ bool CMasternodeGovernance::CanVote(std::string& strErrorRet) const
 {
     if (!masterNodeCtrl.IsActiveMasterNode()){
         strErrorRet = "Only Active Master Node can vote";
+        return false;
     }
 
     auto masterNode = CMasternode{};
     if(!masterNodeCtrl.masternodeManager.Get(masterNodeCtrl.activeMasternode.outpoint, masterNode)){
         strErrorRet = "Failure retrieving Master Node";
+        return false;
     }
 
     if (!masterNode.IsEnabled()){
         strErrorRet = "Only enabled Master Node can vote";
+        return false;
     }
 
     constexpr static auto thirty_days = int{30 * 24 * 60};
     if (masterNode.IsBroadcastedWithin(thirty_days)){
         strErrorRet = "Master Node is not old enough to vote";
-    }
-
-    if (!strErrorRet.empty()){
         return false;
     }
 
--- a/src/mnode/mnode-governance.cpp
+++ b/src/mnode/mnode-governance.cpp
@@ -1,72 +1,81 @@
-// Copyright (c) 2018 The PASTELCoin Developers
+// Copyright (c) 2018-2022 The Pastel Core Developers
 // Distributed under the MIT software license, see the accompanying
-// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+// file COPYING or https://www.opensource.org/licenses/mit-license.php.
 
-#include "main.h"
-#include "key_io.h"
+#include <main.h>
+#include <key_io.h>
+#include <enum_util.h>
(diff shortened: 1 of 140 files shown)


src/mnode/mnode-governance.cpp
```diff
@@ -121,12 +121,38 @@ CAmount CMasternodeGovernance::IncrementTicketPaidAmount(CAmount payment, CGover
...
+        return false;
+    }
+
+    if (!masterNode.IsEnabled()){
+        strErrorRet = "Only enabled Master Node can vote";
+        return false;
+    }
+
+    constexpr static auto thirty_days = int{30 * 24 * 60};
+    if (masterNode.IsBroadcastedWithin(thirty_days)){
```
TimothyBanks, line 143: I'm still a bit unsure how to set up the test to get this condition to pass.

TimothyBanks, line 143: I'm still a bit unsure how to set up the test to get this condition to pass.

src/mnode/mnode-governance.cpp
```diff
@@ -121,12 +121,38 @@ CAmount CMasternodeGovernance::IncrementTicketPaidAmount(CAmount payment, CGover
     return aAmountPaid;
 }
 
+bool CMasternodeGovernance::CanVote(std::string& strErrorRet) const
```
akobrin1, line 124: can make it noexcept...
IsActiveMasterNode,IsEnabled,IsBroadcastedWithin - noexcept
Just make also CMasternodeMan::Get const noexcept- it does not throw any exceptions

src/mnode/mnode-governance.cpp
```diff
@@ -121,12 +121,38 @@ CAmount CMasternodeGovernance::IncrementTicketPaidAmount(CAmount payment, CGover
...
+        strErrorRet = "Failure retrieving Master Node";
+        return false;
+    }
+
+    if (!masterNode.IsEnabled()){
+        strErrorRet = "Only enabled Master Node can vote";
+        return false;
+    }
+
+    constexpr static auto thirty_days_in_seconds = int32_t{30 * 24 * 60 * 60};
```
akobrin1, line 142: you could just declare constexpr static int32_t (or int as the type of parameter in IsBroadcastedWithin function)

src/mnode/mnode-governance.h
```diff
@@ -200,4 +200,7 @@ class CMasternodeGovernance
     void Clear();
 
     void UpdatedBlockTip(const CBlockIndex *pindex);
-};
+
+private:
+    bool CanVote(std::string& strErrorRet) const;
```
akobrin1, line 205: noexcept

akobrin1: Please make it protected & virtual - this way it is easier to write tests. Will just mock it later on.

akobrin1 (changes requested):
Will definitely need to rebase on master - I made some changes in mn-governance.cpp|h files.

a-ok123 (approved):

PR #131, opened 2022-05-14
added intendedFor filed for Sell ticket (no validation yet)

a-ok123:


src/mnode/rpc/tickets-fake.cpp
```diff
@@ -83,8 +83,10 @@ UniValue tickets_fake(const UniValue& params, const bool bSend)
 
         int after = get_number(params[6]);
         int before = get_number(params[7]);
+        
+        string intendedFor = "";
```
akobrin1, line 87: don't really need to set it to ""

src/mnode/rpc/tickets-register.cpp
```diff
@@ -355,8 +355,10 @@ As json rpc:
     string sFundingAddress;
     if (params.size() >= 10)
         sFundingAddress = params[9].get_str();
+    
+    string intendedFor = "";
```
akobrin1, line 359: don't need to set it to ""

akobrin1 (changes requested):

PR #175, opened 2023-04-27
 [PSL-714] Fit it so a node won't reject a block as being invalid just because it was signed by an SN that the node had already banned

akobrin1:
- introduced detailed status enum for getting top mns for the given block:
    SUCCEEDED              : successfully got top masternodes
    SUCCEEDED_FROM_HISTORY : successfully got top masternodes from historical top mn data
    MN_NOT_SYNCED          : masternode is not synced
    BLOCK_NOT_FOUND        : block not found
    GET_MN_SCORES_FAILED   : failed to get masternode scores
    NOT_ENOUGH_MNS         : not enough top masternodes
    HISTORY_NOT_FOUND      : historical top mn data not found
 GetTopMNsForBlock,CalculateTopMNsForBlock,GetMasternodeRanks functions now return this detailed status along with the error message.
 - changed ticket signature validation in case MN synced flag is set:
    - check if there was an error getting top mns for the ticket creator height - return TICKET_VALIDATION_STATE::MISSING_INPUTS along with
	  the detailed error. Block will be cached for the revalidation after some time.
 - enhanced block cache revalidator class:
    - get height of the block if possible from block index map when it is added to the cache
	- sort blocks for revalidation in ascending order of the block height so that blocks with lower height will be revalidated first

### DIFF
--- a/src/mnode/mnode-manager.cpp
+++ b/src/mnode/mnode-manager.cpp
@@ -1,5 +1,5 @@
 // Copyright (c) 2014-2017 The Dash Core developers
-// Copyright (c) 2018-2022 The Pastel Core developers
+// Copyright (c) 2018-2023 The Pastel Core developers
 // Distributed under the MIT/X11 software license, see the accompanying
 // file COPYING or https://www.opensource.org/licenses/mit-license.php.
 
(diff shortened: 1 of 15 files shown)


src/mnode/mnode-manager.cpp
```diff
@@ -168,7 +168,8 @@ void CMasternodeMan::CheckAndRemove(bool bCheckAndRemove)
                     if (vecMasternodeRanks.empty())
                     {
                         int nRandomBlockHeight = GetRandInt(nCachedBlockHeight);
-                        GetMasternodeRanks(vecMasternodeRanks, nRandomBlockHeight);
+                        string error;
```
a-ok123, line 171: error defined here is not used anywhere

a-ok123 (approved):

PR #201, opened 2023-10-03
[PSL-1003] [PSL-1004] [PSL-986] pasteld changes

akobrin1:
[PSL-1003] cNode: change sendtoaddress/sendmany APIs to return change to sender by default
Changed sendtoaddress/sendmany RCP APIs - added optional "change-address" parameter:
sendtoaddress "t-address" amount ("comment" "comment-to" subtractFeeFromAmount "change-address"
sendmany "fromaccount" {"address":amount,...} ( minconf "comment" ["address",...] "change-address"
By default, these functions will send the change (if any) to the original address.

"change-address" (string, optional, default="original") The destination address to send the change to, can be one of the following:
	- "original" - Send the change to the original address (default);
	- "new" - Send the change to the new generated t-address
	- <change t-address> - Send the change to the given Pastel t-address
	
- added python tests for sendtoaddress/sendmany to wallet.py to test all possible "change-address" values:
  default, original, new, t-address, invalid value.

[PSL-986] cNode: benchmark masternode on startup
- in MasterNode mode:
  - check hardware requirements: min CPU count (6) and min RAM (24Gb).
    Exit with error if the machine does not satisfy minimum hardware requirements.
  - perform benchmark and compare with the predefined result. If the machine is not fast enough - exit with an error.
  - added benchmark calibration function
 
[PSL-1004] cNode: remove top 10 MN requirement in ticket signature validation
- removed a check for the signing MN to be in the top 10 MN list for the creation block.
Instead, added a check that MN is actually registered with a valid outpoint.

merged changes:
- check ticket is in stale block or in mempool for NFT, Actions, Collections reg tickets;
 - tickets find - set correct block height for stale blocks (-1);
 - tickets that are not yet downloaded but referred by other tickets added to revalidate queue now

[PSL-1003]: https://pastel-network.atlassian.net/browse/PSL-1003?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PSL-986]: https://pastel-network.atlassian.net/browse/PSL-986?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PSL-1004]: https://pastel-network.atlassian.net/browse/PSL-1004?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


configure.ac
```diff
@@ -3,7 +3,7 @@ AC_PREREQ([2.60])
 define(_CLIENT_VERSION_MAJOR, 1)
```
a-ok123, line 3: should be 2.0

configure.ac
```diff
@@ -3,7 +3,7 @@ AC_PREREQ([2.60])
...
 define(_CLIENT_VERSION_MINOR, 2)
 define(_CLIENT_VERSION_REVISION, 1)
-define(_CLIENT_VERSION_BUILD, 17)
+define(_CLIENT_VERSION_BUILD, 19)
```
a-ok123, line 6: should be 51, so version would be 2.0.0-rc1

src/pasteld.cpp
```diff
@@ -176,12 +183,70 @@ bool AppInit(int argc, char* argv[])
...
+void print_callstack()
+{
+    return;
+}
+#endif
+
+void pasteld_terminate()
+{
+    LogFnPrintf("pasteld_terminate():");
+#ifdef __linux__
```
a-ok123, line 229: don't need it here, i mean #ifdef

a-ok123 (changes requested):

configure.ac
akobrin1: for version 2.0.0-rc1 we need:
define(_CLIENT_VERSION_MAJOR, 2)
define(_CLIENT_VERSION_MINOR, 0)
define(_CLIENT_VERSION_REVISION, 0)
define(_CLIENT_VERSION_BUILD, 25)

src/pasteld.cpp
akobrin1: yep, I've added dummy print_callstack for platforms other than linux

Dicklesworthstone (approved):

PR #247, opened 2024-03-25
[PSL-1143] Do not invalidate block because mnid "is not in the list" if mn list is not finished syncing (mnsync status is not Finished)

akobrin1:
pasteld v2.1.4
- removed the check: find masternode in the list by block signer's mnid and reject the peer that sent this. This can happen if the node is still in initial download mode, but master node list is not initialized or not up-to-date. CheckBlock is already retrieving mnid ticket corresponding to the signer's Pastel ID and checking if this ticket has outpoint defined (mnid ticket).


Dicklesworthstone (dismissed):

src/main.cpp
```diff
@@ -2921,14 +2921,8 @@ bool CheckBlock(
...
-        // so here we need to check that PastelID (mnid) of the signer is registered
-        masternode_info_t mnInfo;
-        if (!masterNodeCtrl.masternodeManager.GetAndCacheMasternodeInfo(block.sPastelID, mnInfo))
-        {
-            strRejectReasonDetails = strprintf("MasterNode with mnid='%s' is not in the list", block.sPastelID);
-            return state.DoS(10, error("%s: %s", __func__, strRejectReasonDetails),
-                REJECT_INVALID, "mn-not-listed", false, strRejectReasonDetails);
-        }
+        // so here we need to check that PastelID (mnid) of the signer was eligible to mine the block
+        // mnid ticket is already checked above
```
a-ok123, line 2925: @akobrin1 wait, you just remove the check altogether?

a-ok123 (approved):

src/main.cpp
akobrin1: yes, I already have a check that mind ticket exists in that function

Dicklesworthstone (approved):

PR #256, opened 2024-07-01
[PSL-1024][PSL-1199][PSL-1215][PSL-1216][PSL-1217][PSL-1218][PSL-1219][PSL-1220][PSL-1221][PSL-1222]

akobrin1:


akobrin1:
pasteld v2.2.4-beta4
[PSL-1215] pasteld: update nlohmann json single-header library
- updated nlohmann json single-header C++ library to v3.11.3 (11/28/2023)

[PSL-1216] pasteld: shutdown optimization
 - refactored pastel shutdown to use condition variable and mutex instead of sleep and simple flag

[PSL-1217] pasteld: ubuntu 22.04 image
Created build image to use for Circle CI builds - based on Ubuntu 22.04 and using GCC 11.

[PSL-1218] pasteld: switch Linux/MSVC to c++20
- changed c++ standard from c++17 to c++20 for Visual Studio and Linux versions.
- changed autoconf macros to check compiler support for c++20 (ax_cxx_compile_stdcxx.m4)

[PSL-1219] pasteld: implement sleep in miner thread
Implemented synchronized sleep for all miner worker threads for the specified number of milliseconds.
- removed 3rdparty cyclicbarrier project used on MSVC version
- used std::barrier implemented in c++20 for synchronized sleep for miner worker threads
- added change priority feature to CServiceThread
- fixed thread priority definitions for Linux
- refactored miner worker threads to use CStoppableServiceThread
- implemented group of miner worker threads on CServiceThreadGroup
- refactored miner thread tests to use parameterized gtests
- introducted new pastel.conf option:
   -gensleepmsecs=<n> : Set the number of milliseconds to sleep for miner threads (default: 100)
- moved all read/write mining settings spreaded across all the project to the CMiningSettings class
- used CStoppableServiceThread condition variable and mutex for all waits in miner thread, this allows to
stop this thread immediately and avoid sleep delays

[PSL-1220] pasteld: use lambda functions for boost slots
Slot functions that was bound by using boost bind and placeholders could not be compiled on c++20. 
These slot functions were changed to simple lambda functions.
- removed boost thread lib from build/link (as all boost threads were changed to std::thread)
- use unordered_map to store signal connections and use it on shutdown while disconnecting signals for validation interfaces

[PSL-1221] pasteld: skip dust check for transactions to burn address
Skipped dust threshold check for transactions to burn address.
- default min relay tx fee changed back to 100 from 30
- corrected python tests

[PSL-1222] pasteld: deadlock on shutdown in COpenConnectionsThread/COpenAddedConnectionsThread
CSemaphoreGrant class was used in COpenConnectionsThread/COpenAddedConnectionsThread.
This class ensures that oubound connections semaphore will be acquired and released in the thread loop scope.
The issue was that this thread loop was using same lock as the CStoppableServiceThread was using in stop() method.
So it was leading to deadlock on shutdown.
Introduced new WorkMutex that was used only in thread loop to avoid deadlock.

[PSL-1024] pasteld: leveldb index for transactions to burn address
scanburntransactions API makes full block scan to search for transactions from specific tracking address to the Pastel Burn transparent address. This consumes a lot of resources (CPU and memory) each time request is executed. 
This functionality has been optimized by creating index in leveldb that will track all transactions to BurnAddress.
On first scanburntransactions API call it makes a full block scan and generate BurnTxIndex.
After that all new blocks should automatically detect transactions to the Burn Address and update BurnTxIndex in leveldb.
Index entries are erased when the block is disconnected from the chain (in case of reorgs).
MasterNode automatically checks if burntxindex is enabled on startup and generates it if not (after MnSync is finished).
BurnTxIndex enabling flag is written to the leveldb and read on startup.
BurnTxIndex allows search by address or by address and starting block height.
- added burn address for the regtest
- added python tests burntxindex.py to test:
   - add 10 new tracking addresses and move some coins there
   - add 5 transactions from the random tracking addresses to the burn address
   - check generation of BurnTxIndex on first scanburntransactions request
   - check result returned by scanburntransactions:
       - for all tracking addresses: mempool only transactions or confirmed transactions 
	   - for the specific tracking address
	   - for the specific tracking address with starting height
   - check that transactions to the burn address are automatically added to the index on ConnectBlock call
   - check that transactions to the burn address are removed in case of chain reorg (used invalidateblock)
   - check that burn tx index stays after node restart
- if destination address parameter in the scanburntransactions API is not burn address, still full block scan is performed

other changes:
  - moved 3rdparty sources to separate subfolder src/extlibs (nlohmann json, utfcpp, scope_guard, signal)
  - fixed an issue in DateTimeStrFormat with local/UTC time
  
http server changes:
  - added listener initialization error handling
  - allow binding to IPv4 endpoint only or to both IPv4/IPv6 endpoints
  - refactored http handlers registration/unregistration, introduced handler groups, added unregister handler group
  - added exception handling for:
     - http listener event loop (libevent)
	 - http server startup
	 - http request processing
	 - http worker queue processing thread
	 - http request handler execution
  - handling of keep-alive connections
  - fixed check for URI max length (1024)
  - default http server timeout increased from 30 to 600 secs
  - unique logging ids for each http worker thread
  - http listener event loop thread implemented on CServiceThread
  - http worker threads implemented on CServiceThreadGroup

Dicklesworthstone (approved):

PR #263, opened 2024-07-30
Utxo sender

a-ok123:
Added new API - `getaddressutxosextra` to include senders (vins) into utxos output

### DIFF
--- a/src/rpc/misc.cpp
+++ b/src/rpc/misc.cpp
@@ -243,7 +243,8 @@ public:
     }
 };
 
-UniValue getUtxosData(vector<pair<uint160, ScriptType>> &vAddresses, uint32_t minHeight, bool includeSender, bool justSendersAddress);
+UniValue getUtxosData(vector<pair<uint160, ScriptType>> &vAddresses, uint32_t minHeight,
+                      const string &senderFilter, bool includeSender, bool justSendersAddress);
 
 UniValue z_validateaddress(const UniValue& params, bool fHelp)
 {
@@ -1075,7 +1076,7 @@ Examples:
     if (!getAddressesFromParams(params, vAddresses))
         throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "Invalid address");
 
-    UniValue utxos = getUtxosData(vAddresses, 0, false, false);
+    UniValue utxos = getUtxosData(vAddresses, 0, "", false, false);
 
     if (!includeChainInfo)
         return utxos;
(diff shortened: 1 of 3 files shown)


src/rpc/misc.cpp
```diff
@@ -1073,13 +1076,117 @@ Result
...
+    rpcDisabledThrowMsg(fInsightExplorer, RPC_API_GETADDRESSUTXOS);
+
+    bool simpleInfo = false;
+    int height = 0;
+    string senderFilter = "";
+    if (params[0].isObject())
+    {
+        UniValue simple = find_value(params[0].get_obj(), "simple");
+        if (!simple.isNull())
+            simpleInfo = simple.get_bool();
```
akobrin1, line 1165: need to use get_bool_value from rpc/rpc-utils.h, that one does all the validations for bool...throws exception

src/rpc/misc.cpp
```diff
@@ -1073,13 +1076,117 @@ Result
...
+        UniValue minHeight = find_value(params[0].get_obj(), "minHeight");
+        if (!minHeight.isNull())
```
akobrin1, line 1167: for minHeight param can use also get_long_number / rpc_check_unsigned_param (like in getblockmininginfo)
`

src/rpc/misc.cpp
```diff
@@ -1073,13 +1076,117 @@ Result
...
+UniValue getUtxosData(vector<pair<uint160, ScriptType>> &vAddresses, uint32_t minHeight, const string &senderFilter,
+                      bool includeSender, bool justSendersAddress) {
     vector<CAddressUnspentDbEntry> vUnspentOutputs;
     for (const auto& it : vAddresses)
     {
         if (!GetAddressUnspent(it.first, it.second, vUnspentOutputs))
             throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "No information available for address");
     }
-    std::sort(vUnspentOutputs.begin(), vUnspentOutputs.end(),
+    sort(vUnspentOutputs.begin(), vUnspentOutputs.end(),
```
akobrin1, line 1189: clang also does not like sort without std::

src/rpc/misc.cpp
```diff
@@ -1100,24 +1210,70 @@ Result
...
+                        }
+                    }
+                }
+                if (!justSendersAddress)
+                    vin.push_back(std::move(in));
+            }
+            if (!senderFilter.empty() && !bFoundSenders) {
+                continue;
+            }
+            output.pushKV("senders", vin);
```
akobrin1, line 1272: std::move(vin)

akobrin1 (changes requested):

akobrin1 (approved):

